PR #13, opened 2014-11-25
Reg up plus

bojh:
Enhance connectivity reliability; fix client object_device examples; add object_location implementation; pure lwm2m_malloc() & lwm2m_free() usage, add client arguments for more convenience local/server port configuration with defaults matching to eclipse leshan.


bojh:
already done!

PR #15, opened 2015-02-03
Cond observe

bojh:
New features:
Support for conditional observe and write attributes. Several attributes set, are logical AND for evaluation.
First block wise implementation covering client response messages.
New connectivity monitoring default object implementation.
Changes: Lifetime handled in core. Additional arguments added to test client.


dnav:
Hi Joerg,
pretty impressive contribution ! Thanks for this.
I'm still reading it but I have one minor comment: could you remove the space/tabulation changes ? It adds a lot of noise to the diff and makes hard to track what the improvements are. Most of the lines of core/observe.c were changed for instance. Usually we use 4 spaces instead of a Tab.

Thanks

jvermillard:
IMO it would be simpler to split it in several pull requests: conditional observe, block transfer, new example objects, etc..

bojh:
Hi David,
yes we will do that, i hope today

bojh:
Hi Julian,
I see it that way. We would/should deliver smaller closed functionalities in the future. This is certainly much easier when combine.

core/packet.c
```diff
@@ -192,93 +182,102 @@ void lwm2m_handle_packet(lwm2m_context_t * contextP,
...
+                    handle_observe_notify(contextP, fromSessionH, message);
+                }
+                else
 #endif
-            {
+#ifdef LWM2M_CLIENT_MODE
+                if (IS_OPTION(message, COAP_OPTION_LOCATION_PATH)) {
+                    multi_option_t *location_path = message->location_path;
+                    if ((location_path->len == 2) && (memcmp(location_path->data, "rd", 2) == 0)) {
+                        handle_registration_response(contextP, fromSessionH, message);
```
dnav: I'm not sure to understand this : are you expecting a registration response in a CON or NON message ?

core/registration.c
```diff
@@ -295,11 +326,40 @@ static int prv_update_registration(lwm2m_context_t * contextP, lwm2m_server_t *
     return 0;
 }
 
+void handle_registration_response(lwm2m_context_t * contextP, void * fromSessionH, coap_packet_t * message)
+{
```
dnav: Shouldn't this code added to handle_registration_reply() function ?

core/packet.c
bojh: “Register” is defined as request/response, and this is defined in COAP-18 2.2. with two possible CON message flows. One with piggy-packed responses (page 12) and one with a separate acknowledge and response-message (page 13). Wakaama only supports piggy-packed responses. Using that, the sender of the result could not ensure, that the result
is already received by the client. Using the separate approach, receiving the ACK for the response message ensures, that the client has already received the response. According to the registration this may cause some problems, when server assume, that the client received the response and therefore send further requests. This patch adds the possibility to process
the registration with separate response also. Currently we work on a more  general transaction to have this possibilities also for other request/response pairs and not only limited to the registration.

core/registration.c
bojh: Yes and even more general, the “transaction” should support the separate responses pattern in general. Currently Achim is working on that …

dnav:
OK.
So actually the issue is in the transaction system which needs two improvements :
- support of the token. For now we only matches replies by sender and MID.
- support the separate response. Upon reception of the ACK we need to determine if the reply is complete. If not the transaction swtches to a new state and waits for the actual data.

With these, this use-case will work. So It may be better to wait for your general solution than having a kind of hack.

Regards

bojh:
Hi David,
so I propose Achim is working currently on that. He should test an finalize this topic, and we will come with an update.

dnav:
Great.
I'm still looking at your changes and I think it would be better to cherry-pick them and have separate pull requests: blockwise, new test objects, compilation warning fixes and write attribute changes are pretty independant.

core/liblwm2m.h
```diff
@@ -332,10 +324,44 @@ typedef enum
...
+{
+    LWM2M_DATATYPE_UNKNOWN = 0,
+    LWM2M_DATATYPE_STRING,
+    LWM2M_DATATYPE_INTEGER,
+    LWM2M_DATATYPE_FLOAT,
+    LWM2M_DATATYPE_BOOLEAN,
+    LWM2M_DATATYPE_OPAQUE,
+    LWM2M_DATATYPE_TIME
+} lwm2m_data_type_t;
+
```
dnav: This could be added to the lwm2m_tlv_t stucture. So no need for a new object callback.

core/liblwm2m.h
```diff
@@ -489,12 +525,35 @@ typedef struct
...
     lwm2m_buffer_send_callback_t    bufferSendCallback;
     void *                          userData;
 } lwm2m_context_t;
 
+typedef uint8_t (*lwm2m_read_callback_t) (uint16_t instanceId, int * numDataP, lwm2m_tlv_t ** dataArrayP, lwm2m_object_t * objectP);
+typedef uint8_t (*lwm2m_write_callback_t) (uint16_t instanceId, int numData, lwm2m_tlv_t * dataArray, lwm2m_object_t * objectP);
+typedef uint8_t (*lwm2m_execute_callback_t) (uint16_t instanceId, uint16_t resourceId, char * buffer, int length, lwm2m_object_t * objectP);
+typedef uint8_t (*lwm2m_create_callback_t) (uint16_t instanceId, int numData, lwm2m_tlv_t * dataArray, lwm2m_object_t * objectP);
+typedef uint8_t (*lwm2m_delete_callback_t) (uint16_t instanceId, lwm2m_object_t * objectP);
+typedef uint8_t (*lwm2m_datatype_callback_t) (int resourceId, lwm2m_data_type_t *resDataType);
```
dnav: I'm very reluctant on adding new object callbacks. This can be achieved by making a Read and having the datatype in the TLV struct

core/attributes.c
```diff
@@ -0,0 +1,470 @@
...
+    //-------------------------------------------------------------------- JH --
+
+    char valStr[50];    //TODO: size enough ? ... malloc?
+    long valLong = 0;
+    // resource attributes are set
+    lwm2m_attribute_data_t * attributeP = NULL;
+
+    if (!LWM2M_URI_IS_SET_RESOURCE(uriP))
+        return COAP_501_NOT_IMPLEMENTED ;
+
```
dnav: here call the object read callback. It will tell you if the resource exists and its datatype.

core/liblwm2m.h
```diff
@@ -332,10 +324,44 @@ typedef enum
...
+  time_t lastTransmission;  /**< holds the time stamp of the last transmission of this object, instance or resource */
+  time_t nextTransmission;  /**< holds the time for the next scheduled transmission */
+  lwm2m_uri_t uri;          /**< holds the URI of the object/instance/resource the attributes are valid for */
+  char* minPeriod;       /**< minimal update period in seconds, the object/instance/resource must not be sent more frequent */
+  char* maxPeriod;       /**< maximum update period, the object/instance/resource must be sent at least after this period, even if no change happend */
+  char* greaterThan;     /**< the resource has to be sent if this threshold is reached */
+  char* lessThan;        /**< the resource has to be sent if this threshold is underrun */
+  char* step;            /**< the resource has to be sent if it is changed by the given step */
+  char* oldValue;        /**< holds the last sent value of the resource if the resource is a numerical resource */
+  lwm2m_data_type_t resDataType;
```
dnav: Why are you storing the attributes as char* and not time_t and float ? This adds unnecessary convertions.

core/attributes.c
bojh: good idea, getting value and type (see previous question) !

core/liblwm2m.h
bojh: Ok, the benefit of is this solution is: The NULL pointer init shows unset mode. The question is valid, we also discussed that topic. The alternative would be a union construction with an additional set flag. May be better but a little bit more sophisticated.

core/observe.c
```diff
@@ -259,49 +261,158 @@ void cancel_observe(lwm2m_context_t * contextP,
...
+    
+    observedP = contextP->observedList;
+
+    while (observedP != NULL)
+    {
+        notification_data_t data;
+        memset(&data, 0, sizeof(data));
+        lwm2m_watcher_t * watcherP = observedP->watcherList;
+        while (watcherP != NULL) {
+          lwm2m_attribute_data_t * attributeData = lwm2m_getAttributes(watcherP->server, &(observedP->uri));
```
dnav: Maybe you should add in the lwm2m_watcher_t a pointer to the revelant lwm2m_attribute_data_t. This would avoid parsing the list each time.

bojh:
Hi David, 
to your cherry picking proposal: 
Feel free - prioritize and peck the most of it. May be this is the best.

core/observe.c
bojh: OK - is an argument - I should have a look to that

bojh:
Hi David,
to your cherry picking argument: Feel free priorise pick

core/management.c
```diff
@@ -55,7 +55,8 @@ coap_status_t handle_dm_request(lwm2m_context_t * contextP,
                                 lwm2m_uri_t * uriP,
                                 void * fromSessionH,
                                 coap_packet_t * message,
-                                coap_packet_t * response)
+                                coap_packet_t * response,
+                                bool* notify_change)
 {
```
dnav: About this notify_change behavior : IMHO it should always be the object's responsability to trigger notify through lwm2m_resource_value_changed() API. It would prevent having two notifications sent for the same event.

dnav:
So I just created two new pull requests for you to review. One containing your various fixes. The other containing your changes to the client and objects. 
I'm sorry but it is very tricky to split the write attributes and the blockwise.
For the registration handling, I'm waiting for the token implementation in transactions.

core/transaction.c
```diff
@@ -234,11 +234,52 @@ void transaction_handle_response(lwm2m_context_t * contextP,
                 // So we resend transaction that were denied for authentication reason.
                 if (message->code != COAP_401_UNAUTHORIZED || transacP->retrans_counter >= COAP_MAX_RETRANSMIT)
```
dnav: For the record, this HACK will be removed for a better one:
When a client receives a CON packet from a server in REG_PENDING_STATE, it will silently ignore it. Thus the server will retransmit the CON and it should arrive after the registration ACK.

bojh: Hi David,
shortly comment: tricky I know ;-) Transaction is work in process coming next week by my colleague. Splitting is OK but I will review it on Monday.

bojh:
From the current point of view I would propose to close this PR? - Please vote!
My colleague is working on an complete and separated "block-wise" and "transaction" solution PR.
From my side I will be one week offline for skying.

jvermillard:
+1 for more PR
Have fun! try avoid breaking your shoulder :)

bojh:
OK - I will close this PR - Have fun too!

PR #16, opened 2015-02-06
Manual Integration: Merge changes to the test client and objects from Bo...

dnav:
...sch SI in bsinno:cond_observe branch.

### DIFF
--- a/tests/client/lwm2mclient.c
+++ b/tests/client/lwm2mclient.c
@@ -385,7 +385,7 @@ int main(int argc, char *argv[])
     const char * server = "localhost";
     const char * serverPort = LWM2M_STANDARD_PORT_STR;
     char * name = "testlwm2mclient";
-    int livetime = 300;
+    int lifetime = 300;
     int batterylevelchanging = 1;
     time_t reboot_time = 0;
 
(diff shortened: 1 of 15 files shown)


bojh:
Hi David,
from my short review some remarks:
1. minor bug: in tests/client/object_connectivity_moni.c in line 362: should be
strcpy(myData->ipAddresses[1], VALUE_IP_ADDRESS_2);
2. Error related to our original source: A read request from leshan-standalone from /4/0 does now generate an alert: "Unable to read instance /4/0 for testlwm2mclient : 500 [Unable to decode tlv.] ([/4/0])"
...My first idea was: This belongs to the now missing "blockwise" functionality, because the response is truncated could not be confirmed because a double-check with the original cond_observe PR does have the same problem. So I can't drill it down at the moment.

tests/client/lwm2mclient.c
```diff
@@ -390,7 +456,7 @@ int main(int argc, char *argv[])
     }
 
     int serverId = 123;
-    objArray[3] = get_server_object(serverId, "U", 300, false);
+    objArray[3] = get_server_object(serverId, "U", livetime, false);
```
jvermillard: livetime=> lifetime

bojh: ... still a live :-) 
lifetime is much better!
I expect David will change it - right ?

jvermillard:
yes the payload is truncated to 128 in er-coap-13/er-coap-13.c
set_payload(..) function.
So yes lack of block transfer is hurting here

bojh:
The alternative would be to shrink the multi-instance resources in the example to fit in one block-size, having a working still working object?

jvermillard:
or raise the 128 bytes limit which is really low for usual Internet MTU

dnav:
Hi,
I just fixed the two minor issues. What about letting the "Read /4/0" issue as a test for the block transfer integration ?

jvermillard:
+1

bojh:
my new PR contains modified Connectivity Montiring implementation with reduces resource instances fitting to the current 128 Byte payload ... up to the blockwise implementation.

PR #18, opened 2015-02-24
Bugfixes

boaks:
Bugfixes:
- some (potential) memory leaks
- initial observe request
- payload size in test-client and -server
- tlv decoding values with larger lengths (e.g. between 128 and 255)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,11 +11,14 @@ Makefile
 
 .cproject
 .project
+.settings
 
 CMakeCache.txt
 CMakeFiles
 Debug
 Release
+DebugServer
+ReleaseServer
 cmake_install.cmake
 core/CMakeCache.txt
 core/CMakeFiles
--- a/README
+++ b/README
@@ -46,7 +46,7 @@ In the any directory, run the following commands:
     make
     ./lwm2mserver
 
-The lwm2mserver listens on UDP port 5684. It features a basic command line
+The lwm2mserver listens on UDP port 5683. It features a basic command line
 interface. Type 'help' for a list of supported commands.
 
 
@@ -75,6 +75,6 @@ The lwm2mclient features four LWM2M objects:
            test |  1 |    R/W     |    No     |    Yes    | Integer | 0-255 |
            exec |  2 |     E      |    No     |    Yes    |         |       |
     
-The lwm2mclient opens udp port 5683 and tries to register to a LWM2M Server at
-127.0.0.1:5684. It features a basic command line interface. Type 'help' for a
+The lwm2mclient opens udp port 56830 and tries to register to a LWM2M Server at
+127.0.0.1:5683. It features a basic command line interface. Type 'help' for a
 list of supported commands.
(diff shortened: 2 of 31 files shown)


boaks:
"I can see the memory leak but I don't get your fix. Why call coap_set_payload() twice ?"
Your right, one is enough :-). 

"Why not test the return of handle_observe_request() and act accordingly ?"
Do you use a "memory sanitizer" tool to detect leaks?
As far as my tools signals, "object_read" allocates the buffer. If you don't set this buffer as response, the pointer gets lost and this will cause a "memory leak".
So, if the server request an observation, the buffer gets allocated. If then handle_observe_request() returns an error (e.g. see above COAP_401_UNAUTHORIZED, when update registration is pending), the buffer is not set as payload and therefore gets lost as memory leak.

boaks:
"From a security point-of-view, it is better to discard requests from server in Update Pending State."

I didn't find something about that special security problem during update registration in the LwM2M specification. I even didn't find a hint for special LwM2M-client side processing during that period in the LwM2M specification. So could you help me out, pointing me to information, which describes this risc and the special processing rules?

Be aware, that the "update registration" may cross the "observe" so the server has no chance to detect the situation ahead. And it would get rather complex, to process the COAP_401_UNAUTHORIZED depending on a pending update registration. 
(I guess, currently no server will handle this properly.)

dnav:
Regarding the handle_observe_request() memory leak:
if you set the buffer as the payload, it will be returned in the response containing the 4.01 code, creating a security hole. My idea was to change the code to:
 result = object_read(contextP, uriP, &buffer, &length);
 if (result == COAP_205_CONTENT)
 {
       if (IS_OPTION(message, COAP_OPTION_OBSERVE))
       {
            result = handle_observe_request(contextP, uriP, fromSessionH, message, response);
       }
       if (result == COAP_205_CONTENT)
       {
            coap_set_payload(response, buffer, length);
            // lwm2m_handle_packet will free buffer
        }
+      else
+      {
+            lwm2m_free(buffer);
+      }
  }

boaks:
"if you set the buffer as the payload, it will be returned in the response containing the 4.01 code, creating a security hole."

Very good point! So free it explicitly is the better choice.

dnav:
About the observation while updating the registration:
There are no advice on how to handle requests during the update in the TS. I was just expressing my concern. But upon close examination, I actually can't see how it can hurt. So let's go with your change.

boaks:
"There is currently a mismatch between CoAP Observe spec and LWM2M TS on this. Let's follow LWM2M TS until this is sorted out. But note 2.04 is a valid code for a notify message."

"OMA-DM-LightweightM2M-2015-0007R01-CR_Erratas_Complement" mentions the 2.04 as error and fixed it to 2.05 (8.2.5, page 3). Don't know, why the TS wasn't updated also. But it's also Ok to wait for a TS update.

boaks:
OMA-TS-LightweightM2M-V1_0-20150228-D, 8.2.5, page 65.
Now the 2.05 is in the specification (at least n the draft).

dnav:
I saw this.
Please change the fix for the memory leak in handle_observe_request() and I'll merge your PR.

boaks:
Changed.

dnav:
Could you rebase to eclipse:master ? There is a small conflict in lwm2mclient.c.
I moved some curly braces. Sorry for that.

boaks:
Rebased.

PR #19, opened 2015-02-26
Di bootstrap

prieux:
Implementation of client initiated bootstrap in "NO SEC" mode.
Bootstrap sequence can be launched at startup given the "bootstrap" parameter (see README).
Also when registered to a DM server, in case of registration update failure, a new bootstrap sequence is launched and client should register to the newly provisioned server.

### DIFF
--- a/core/bootstrap.c
+++ b/core/bootstrap.c
@@ -1,0 +1,188 @@
+/*******************************************************************************
+ *
+ * Copyright (c) 2013, 2014 Intel Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * The Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * Contributors:
+ *    Pascal Rieux - Please refer to git log
+ *
+ *******************************************************************************/
+
+/*
+ Copyright (c) 2013, 2014 Intel Corporation
+
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+
+     * Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+     * Redistributions in binary form must reproduce the above copyright notice,
+       this list of conditions and the following disclaimer in the documentation
+       and/or other materials provided with the distribution.
+     * Neither the name of Intel Corporation nor the names of its contributors
+       may be used to endorse or promote products derived from this software
+       without specific prior written permission.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+ OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ THE POSSIBILITY OF SUCH DAMAGE.
+
+ David Navarro <david.navarro@intel.com>
+
+*/
+
+#include "internals.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+#ifdef LWM2M_CLIENT_MODE
+#define PRV_QUERY_BUFFER_LENGTH 200
+
+static int prv_getBootstrapQuery(lwm2m_context_t * context,
+        char * buffer,
+        size_t length) {
+    int index = 0;
+
+    index = snprintf(buffer, length, "?ep=%s", context->endpointName);
+    if (index <= 1) {
+        return 0;
+    }
+    return index;
+}
+
+static void prv_handleBootstrapReply(lwm2m_transaction_t * transaction, void * message) {
+    lwm2m_server_t * targetP;
+    
+    LOG("[BOOTSTRAP] Handling bootstrap reply...\r\n");
+    lwm2m_context_t * context = (lwm2m_context_t *)transaction->userData;
+    coap_packet_t * coapMessage = (coap_packet_t *)message;
+    if ((NULL != coapMessage) && (coapMessage->type == COAP_TYPE_ACK)) {
+        handle_bootstrap_ack(context, coapMessage, NULL);
+    }
+    else {
+        bootstrap_failed(context);
+    }
+}
+
+// start a device initiated bootstrap
+int lwm2m_bootstrap(lwm2m_context_t * context) {
+    char query[PRV_QUERY_BUFFER_LENGTH];
+    int query_length = 0;
+    lwm2m_transaction_t * transaction = NULL;
+
+    query_length = prv_getBootstrapQuery(context, query, sizeof(query));
+    if (query_length == 0) {
+        return INTERNAL_SERVER_ERROR_5_00;
+    }
+
+    // find the first bootstrap server
+    lwm2m_server_t * bootstrapServer = context->bootstrapServerList;
+    if (bootstrapServer != NULL) {
+        if (bootstrapServer->sessionH == NULL) {
+            bootstrapServer->sessionH = context->connectCallback(bootstrapServer->shortID, context->userData);
+        }
+        if (bootstrapServer->sessionH != NULL) {
+            LOG("[BOOTSTRAP] Bootstrap session starting...\r\n");
+            transaction = transaction_new(COAP_POST, NULL, context->nextMID++, ENDPOINT_SERVER, (void *)bootstrapServer);
+            if (transaction == NULL) {
+                return INTERNAL_SERVER_ERROR_5_00;
+            }
+            coap_set_header_uri_path(transaction->message, "/"URI_BOOTSTRAP_SEGMENT);
+            coap_set_header_uri_query(transaction->message, query);
+            transaction->callback = prv_handleBootstrapReply;
+            transaction->userData = (void *)context;
+            context->transactionList = (lwm2m_transaction_t *)LWM2M_LIST_ADD(context->transactionList, transaction);
+            if (transaction_send(context, transaction) == 0) {
+                bootstrapServer->mid = transaction->mID;
+                LOG("[BOOTSTRAP] DI bootstrap requested to BS server\r\n");
+                // Backup the object configuration in case of network or bootstrap server failure
+                lwm2m_backup_objects(context);
+                lwm2m_deregister(context);
+            }
+        }
+        else {
+            LOG("No bootstrap session handler found\r\n");
+        }
+    }
+    return NO_ERROR;
+}
+
+void handle_bootstrap_ack(lwm2m_context_t * context,
+        coap_packet_t * message,
+        void * fromSessionH) {
+    if (COAP_204_CHANGED == message->code) {
+        context->bsState = BOOTSTRAP_PENDING;
+        LOG("[BOOTSTRAP] Received ACK/2.04, Bootstrap pending, waiting for DEL/PUT from BS server...\r\n");
+        reset_bootstrap_timer(context);
+        delete_bootstrap_server_list(context);
+    }
+    else {
+        bootstrap_failed(context);
+    }
+}
+
+void bootstrap_failed(lwm2m_context_t * context) {
+    context->bsState = BOOTSTRAP_FAILED;
+    LOG("[BOOTSTRAP] Bootstrap failed\r\n");
+    lwm2m_restore_objects(context);
+}
+
+void reset_bootstrap_timer(lwm2m_context_t * context) {
+    struct timeval currentTime;
+
+    lwm2m_gettimeofday(&currentTime, NULL);
+    context->bsStart.tv_sec = currentTime.tv_sec;
+}
+
+void lwm2m_update_bootstrap_state(lwm2m_context_t * context,
+        uint32_t currentTime,
+        struct timeval * timeout) {
+    if (context->bsState == BOOTSTRAP_REQUESTED) {
+        context->bsState = BOOTSTRAP_CLIENT_HOLD_OFF;
+        context->bsStart.tv_sec = currentTime;
+        LOG("[BOOTSTRAP] Bootstrap requested at: %u, now waiting during ClientHoldOffTime...\r\n",
+                context->bsStart.tv_sec);
+    }
+    if (context->bsState == BOOTSTRAP_CLIENT_HOLD_OFF) {
+        lwm2m_server_t * bootstrapServer = context->bootstrapServerList;
+        if (bootstrapServer != NULL) {
+            // get ClientHoldOffTime from bootstrapServer->lifetime
+            // (see objects.c => object_getServers())
+            if ((currentTime - context->bsStart.tv_sec) > bootstrapServer->lifetime) {
+                lwm2m_bootstrap(context);
+            }
+        }
+        else {
+            bootstrap_failed(context);
+        }
+    }
+    if (context->bsState == BOOTSTRAP_PENDING) {
+        if ((currentTime - context->bsStart.tv_sec) > timeout->tv_sec) {
+            // Time out and no error => bootstrap OK
+            LOG("\r\n[BOOTSTRAP] Bootstrapped at: %u (difftime: %u s)\r\n",
+                    currentTime, currentTime - context->bsStart.tv_sec);
+            context->bsState = BOOTSTRAPPED;
+            delete_transaction_list(context);
+            delete_server_list(context);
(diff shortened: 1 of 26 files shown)


jvermillard:
pascalrieux64@gmail.com have a CLA in the database, so it's fine for merging

jvermillard:
It was tested versus leshan bootstrap server

core/liblwm2m.h
```diff
@@ -16,7 +16,11 @@
  *    Simon Bernard - Please refer to git log
  *    Toby Jaffey - Please refer to git log
  *    Julien Vermillard - Please refer to git log
+<<<<<<< HEAD
  *    Bosch Software Innovations GmbH - Please refer to git log
+=======
+>>>>>>> 921c42433134893b75146a7dd5d75af756bb6e1c
+ *    Pascal Rieux - Please refer to git log
```
jvermillard: please fix the header :)

core/internals.h
```diff
@@ -14,7 +14,12 @@
  *    David Navarro, Intel Corporation - initial API and implementation
  *    Fabien Fleutot - Please refer to git log
  *    Toby Jaffey - Please refer to git log
+<<<<<<< HEAD
  *    Bosch Software Innovations GmbH - Please refer to git log
+=======
+>>>>>>> 921c42433134893b75146a7dd5d75af756bb6e1c
```
jvermillard: same :)

core/bootstrap.c
```diff
@@ -0,0 +1,188 @@
...
+
+    index = snprintf(buffer, length, "?ep=%s", context->endpointName);
+    if (index <= 1) {
+        return 0;
+    }
+    return index;
+}
+
+static void prv_handleBootstrapReply(lwm2m_transaction_t * transaction, void * message) {
+    lwm2m_server_t * targetP;
```
jvermillard: unused variable?

boaks:
Hi Pascal,

would it be possible , to try to contribute again?

With "Failed --- The pull request did not pass Eclipse validation. The following users do not have Eclipse account" it would be hard discussion with some legal departments, even with Juliens clarification.

jvermillard:
@boaks don't trust the verification script, it's buggy :) some days it just doesn't trigger at all.
it's really just an helper.

https://projects.eclipse.org/user/cla/validate is the right tool for checking.
Anyway pascal is a coworker of me @ sierra wireless. Sierra as an eclipse member have an agreement covering all the employee and pascal signed the CLA (which you can verify with the CLA tool.

boaks:
Hi Julien, thanks for the clarification.

dnav:
Hi,

Thank you Pascal for your contribution.

I can see you are using a gmail adress for signing the code and your github account is not linked in any way to Sierra Wireless. So it is not clear if you are contributing as an indivudual or as a Sierra Wirless employee. Thus the confusion.

(of course Julien I trust your words ;))

prieux:
Hi David,

You are welcome. About the link with Sierra Wireless, that was the purpose of the alternate email I added while completing the CLA. Unfortunately it didn't work, see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=460955.
Anyway, I'll fix the copyright issue.
Thanks

prieux:
The goal of backuping security and server objects is to handle broken network connection while bootstrapping. And the backup itself is handled by the copyFunc inside the objects while the client only maintains a list of the backuped objects. In the bootstrap sequence, the backup is done as soon as the bootstrap transaction starts.

dnav:
Regarding the backup:
I understand the use-case. My point is more that if you want this feature, it should be done by the client application and not in the core.

core/internals.h
```diff
@@ -74,8 +76,11 @@
 #define URI_REGISTRATION_SEGMENT_LEN    2
 #define URI_BOOTSTRAP_SEGMENT           "bs"
 #define URI_BOOTSTRAP_SEGMENT_LEN       2
+#define URI_DELETE_ALL_SEGMENT          "/"
+#define URI_DELETE_ALL_SEGMENT_LEN      1
```
dnav: Are you sure this is working ?
Because in CoAP the "/" URI is expressed as having no Uri-Path option in the message.

jvermillard: I think the bug is coming from leshan:
https://github.com/eclipse/leshan/blob/master/leshan-server-cf/src/main/java/org/eclipse/leshan/server/californium/impl/BootstrapResource.java#L102

We also need to fix that for leshan bootstrap server

boaks: I even didn't understand the LWM2M bootstrap interface in this issue "delete all".
The bootstrap interface doesn't provide read function, so the only option to get to a defined point is to delete all. So I don't see, why this "delete all" is left to the bootstrap serve instead of being mandatory for a bootstrap.

dnav: The bootstrap server is the only one able to access the security object. If you are adding a new LWM2M server to your network, you may want to configure your clients with this new server without factory-resetting them.

jvermillard:
About the backup, I agree doing it in the framework in-memory is not great. But if we want to delegate it to the user code, maybe we need also to  propagate start/end of bootstrap event, so the end-user can implement his fallback strategy in case on incomplete of incoherent bootstrap ?

dnav:
@jvermillard, this is why the lwm2m_context_t structure is defined in liblwm2m.h. Between each call to lwm2m_step(), the user code can check the current state of the stack and act accordingly.

core/objects.c
```diff
@@ -189,35 +189,67 @@ coap_status_t object_write(lwm2m_context_t * contextP,
...
-        size = 1;
-        tlvP = lwm2m_tlv_new(size);
-        if (tlvP == NULL) return COAP_500_INTERNAL_SERVER_ERROR;
-
-        tlvP->flags = LWM2M_TLV_FLAG_TEXT_FORMAT | LWM2M_TLV_FLAG_STATIC_DATA;
-        tlvP->type = LWM2M_TYPE_RESSOURCE;
-        tlvP->id = uriP->resourceId;
-        tlvP->length = length;
-        tlvP->value = (uint8_t*)buffer;
+    if (NULL == targetP) {
```
dnav: If you did not find a lwm2m_object_t handling this object ID, everything below will fail.

jvermillard:
Ok, it's fine, so it's up to the use to monitor bootstrap start & stop and check the configuration, looking at global state in lwm2mm_context_t. So maybe we can implement config backup in the testclient.
My concern is to be sure a user don't miss the fact it's mandatory to have this backup mechanism, due to the nature of the protocol.

core/objects.c
prieux: Wrong. Code starting line 245 is still executed.

dnav: I meant the code insid the if (NULL == targetP) : the object_create() will fail each time.
You should change to:
```C
if (NULL == targetP)
{
    bootstrap_failed(contextP);
    return COAP_404_NOT_FOUND;
}
```

prieux: Sorry you're right. I'll fix that.

prieux: done removing useless code.

core/uri.c
```diff
@@ -121,6 +122,14 @@ lwm2m_uri_t * lwm2m_decode_uri(multi_option_t *uriPath)
...
         return uriP;
     }
+    else if (URI_DELETE_ALL_SEGMENT_LEN == uriPath->len
+     && 0 == strncmp(URI_DELETE_ALL_SEGMENT, uriPath->data, uriPath->len))
+    {
+        uriP->flag |= LWM2M_URI_FLAG_DELETE_ALL;
+        uriPath = uriPath->next;
+        if (uriPath != NULL) goto error;
+        return uriP;
+    }
```
dnav: This part still need to be fixed. Thanks.

jvermillard: I pused the fix in leshan: https://github.com/eclipse/leshan/commit/aa1e4c492519b0ee378044c7a0840e6e4cfca732

core/packet.c
```diff
@@ -109,31 +110,46 @@ static coap_status_t handle_request(lwm2m_context_t * contextP,
     coap_status_t result = NOT_FOUND_4_04;
 
     uriP = lwm2m_decode_uri(message->uri_path);
-    if (uriP == NULL) return BAD_REQUEST_4_00;
-
-    switch(uriP->flag & LWM2M_URI_MASK_TYPE)
-    {
+    if (uriP == NULL) {
```
dnav: uriP can be NULL is case of invalid URI, thus a Delete on /2/toto would be interpreted as a Delete on /.
You should keep a special flag for / and return it when the uriPath is empty.

dnav:
After some testing, I'm not sure we need a bootstrap state.
The TS §5.2.2.3 and 5.2.2.4 state:
> the LWM2M Client MUST purge the LWM2M Bootstrap Server Account after the timeout value Bootstrap Server Account Timeout, if defined, or immediately after successful bootstrapping if the timeout value is not defined.

So if the client has a bootstrap server account, the bootstrap operations are possible.

jvermillard:
@dnav this is a recent addition to the draft and IMO it's broken.
Ok let's purge bootstrap server accounts, but how are you going to change your X509 certificate which is having an expiring date? Credentials stay forever? already reported this issue to some OMA representatives.

dnav:
@jvermillard, another good point. I opened this issue against the OMA public review:
https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/39

core/packet.c
prieux: Do you mean in the very special case of bootstrapping or in a more generic case? If it's in a generic case, I would prefer create a new branch for that and not mix a related fix in the di_bootstrap branch (so this one could be merged without having to rebase too often onto master...)

dnav: I meant in this special case. You need to check if the delete is on "/". Currently you do a handle_delete_all() for any invalid URI.
You need a special handling of "/" URI in lwm2m_decode_uri() setting a fleg that you'll test before branching below. The case uriP==NULL will return a Bad Request code.

jvermillard:
IMO we should not support the BS credential purge until the spec looks fine.

@dnav what is missing for your point-of-view for merging?

dnav:
Indeed. IMHO, the boostrap server is required through the whole lifetime of the client. I have use-cases where a LWM2M server is added to an already running LWM2M system. These use-cases are no longer possible if the Bootstrap server is purged.

@jvermillard, before merging, I'd like the point in packet.c line 113 to be fixed. In current state a Delete on an invalid URI would be treated as a Delete on /.
Also, I'm not sure the boostrap state is needed as the boostrap server may send commands at any time during the client's lifetime.

eclipsewebmaster:
:white_check_mark: 
#### Eclipse validation results:
All committers passed Eclipse CLA and Signed-off-by validation.

bojh:
Getting familiar with the bootstrap use case, I found that the TS-20150228-D does reference (page 91,128) a Security object "Short Server ID" value=0 in the case of an bootstrap setting. This would imply a modification in objects.c line 555:
554        if (0 == lwm2m_tlv_decode_int(tlvP + 2, &value)
555         || value < ((isBootstrap)?0:1) || value > 0xFFFF)     // 0: forbidden for none-bootstrap
556        {

dnav:
Hi @prieux,

please remove the backup/restore features from the core as discussed previously. The way the bootstrap sequence is defined currently in LWM2M TS, the bootstrap can only occur at client boot-time. So backup/restore are not mandatory.
Also, the way it is implemented in the client callback, it can not work as you do not copy the internal structures of the plugins.

Regards

prieux:
Hi @dnav, please have a look at commits 5ba5b55 and c5dbfbc, backup/restore has already been moved from core to client.
BR

core/liblwm2m.c
```diff
@@ -57,8 +58,10 @@
 
 
 lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback,
-                             lwm2m_buffer_send_callback_t bufferSendCallback,
-                             void * userData)
+        lwm2m_buffer_send_callback_t bufferSendCallback,
+        backup_objects_callback_t backupObjectsCallback,
```
dnav: Here

core/liblwm2m.c
```diff
@@ -68,6 +71,12 @@ lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback,
 #ifdef LWM2M_CLIENT_MODE
     if (NULL == connectCallback)
         return NULL;
+    if (NULL == backupObjectsCallback) {
+        return NULL;
+    }
+    if (NULL == restoreObjectsCallback) {
+        return NULL;
+    }
```
dnav: Here

core/liblwm2m.c
```diff
@@ -79,70 +88,130 @@ lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback,
         contextP->userData = userData;
         srand(time(NULL));
         contextP->nextMID = rand();
+#ifdef LWM2M_CLIENT_MODE
+        contextP->backupObjectsCallback = backupObjectsCallback;
+        contextP->restoreObjectsCallback = restoreObjectsCallback;
+#endif
```
dnav: Here

core/liblwm2m.c
```diff
@@ -79,70 +88,130 @@ lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback,
...
+                    }
+                }
+                /*
+                 * Let's assume that if we want to delete ALL objects
+                 * we need to free objects as well
+                 */
+                if (!securityAndServerOnly) {
+                    lwm2m_free(objectList[i]);
+                    objectList[i] = NULL;
+                }
```
dnav: Please don't free. Its the job of closFunc() callback.

core/liblwm2m.h
```diff
@@ -473,18 +485,28 @@ typedef void * (*lwm2m_connect_server_callback_t)(uint16_t serverID, void * user
 // The session handle MUST uniquely identify a peer.
 typedef uint8_t (*lwm2m_buffer_send_callback_t)(void * sessionH, uint8_t * buffer, size_t length, void * userData);
 
+typedef struct _lwm2m_context_ lwm2m_context_t;
 
-typedef struct
+typedef void (*backup_objects_callback_t)(lwm2m_context_t * context);
+typedef void (*restore_objects_callback_t)(lwm2m_context_t * context);
+
```
dnav: Here

core/liblwm2m.h
```diff
@@ -473,18 +485,28 @@ typedef void * (*lwm2m_connect_server_callback_t)(uint16_t serverID, void * user
...
     int    socket;
 #ifdef LWM2M_CLIENT_MODE
+    lwm2m_bootstrap_state_t bsState;
+    struct timeval      bsStart;
     char *              endpointName;
     char *              msisdn;
     lwm2m_server_t *    bootstrapServerList;
     lwm2m_server_t *    serverList;
     lwm2m_object_t **   objectList;
+    lwm2m_object_t **   objectListBackup;
```
dnav: Here

core/liblwm2m.h
```diff
@@ -473,18 +485,28 @@ typedef void * (*lwm2m_connect_server_callback_t)(uint16_t serverID, void * user
...
     uint16_t            numObject;
+    uint16_t            numObjectBackup;
     lwm2m_observed_t *  observedList;
+    backup_objects_callback_t backupObjectsCallback;
+    restore_objects_callback_t restoreObjectsCallback;
```
dnav: Here

core/liblwm2m.h
```diff
@@ -519,6 +546,15 @@ int lwm2m_configure(lwm2m_context_t * contextP, char * endpointName, char * msis
 // create objects for known LWM2M Servers.
 int lwm2m_start(lwm2m_context_t * contextP);
 
+// manage a bootstrap session
+int lwm2m_bootstrap(lwm2m_context_t * contextP);
+
+// backup objects (0 and 1) configuration and content (during bootstrap)
+void lwm2m_backup_objects(lwm2m_context_t * contextP);
```
dnav: Here

core/liblwm2m.h
```diff
@@ -519,6 +546,15 @@ int lwm2m_configure(lwm2m_context_t * contextP, char * endpointName, char * msis
...
+
+// restore objects (0 and 1) configuration and content (in case of bootstrap failure)
+void lwm2m_restore_objects(lwm2m_context_t * contextP);
```
dnav: Here

dnav:
@prieux I tagged some places in core still referencing the backup/restore feature.

dnav:
@bojh This change is needed if we consider that a bootstrap server can create another bootstrap server account. With the current bootstrap sequence, I can't see this happening.

bojh:
My issue was directed by the example on page 107 table 28, where the one and only bootstrap security object instance references  "Short Server ID" = 0; In my understanding  e.g. a factory-BS which could regularly create a productive BS, and this could be changed over time or devices customer charge.

bojh:
@prieux, Additional suggestions: 
liblwm2m.c line 102:
fprintf(stderr, "Usage: lwm2mclient [..[localPort] server] serverPort] name] lifetime] batChg] bootstrap\r\n");
liblwm2m.c line 102: correct it to:
            {"bootstrap", "Initiate a DI bootstrap process", NULL, prv_initiate_bootstrap, NULL},

boaks:
@dnav "This change is needed if we consider that a bootstrap server can create another bootstrap server account. With the current bootstrap sequence, I can't see this happening."

Is it possible, that you point us to the parts in the TS, which indicates that?

I found in the TS-20150228, 5.2.1. "Bootstrap Information" two variants, the
"LWM2M Server Bootstrap Information" and the "LWM2M Bootstrap Server Bootstrap Information".
Though section 5.2.2.3 use the term "Bootstrap Information" for the data written, I can't see that this is limited to the "LWM2M Server Bootstrap Information" and excludes the "LWM2M Bootstrap Server Bootstrap Information". And E.1. does also not mention, that a security object for a "Bootstrap Server" can not be changed by a Bootstrap Server.
May be, this is just a miss- or over-interpretation from my side, and the "LWM2M Bootstrap Server Bootstrap Information" can not be changed by a Bootstrap server.
If so, I would purpose to add a problem report to the OMA to clarify the wording.

dnav:
@boaks Actually bojh is right. 
The bootstrap server can indeed write or modify a bootstrap server account. My bad.

prieux:
@dnav moved last references to backup/restore from core to client

boaks:
In "lwm2mclient.c" line 602, a array of 6 char is defined.

    "char bootstrapRequested[6];"

but, if the expected term "bootstrap" is given (strcpy to the buffer), this may overflow the 6 char.

boaks:
In "bootstrap.c" function "update_bootstrap_state" the "transactions" and "servers" are deleted, when the bootstrap finished. Are there reasons, why the "bootstrap servers" are not deleted also? Servers and bootstrap servers are "recreated" in "object_getServers" so I'm wondering, why they are not deleted.

dnav:
These changes are now merged in a new branch named "new_bootstrap" created from current master. Please continue work there.

PR #20, opened 2015-02-27
Single instance test objects now declare their instance.

dnav:
This fixes the registration payload being /1/0, /3, /4, /5, /6, /1024/10, /1024/11, /1024/12
instead of /1/0, /3/0, /4/0, /5/0, /6/0, /1024/10, /1024/11, /1024/12

### DIFF
--- a/tests/client/object_connectivity_moni.c
+++ b/tests/client/object_connectivity_moni.c
@@ -43,8 +43,6 @@
 #include <ctype.h>
 
 // related to TS RC 20131210-C (ATTENTION changes in -D!)
-// Object Id
-#define OBJ_DEVICE_ID                   4
 // Resource Id's:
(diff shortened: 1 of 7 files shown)


tests/client/object_connectivity_moni.c
```diff
@@ -334,11 +332,25 @@ lwm2m_object_t * get_object_conn_m()
         memset(connObj, 0, sizeof(lwm2m_object_t));
 
         /*
-         * It assign his unique ID
-         * The 3 is the standard ID for the mandatory object "Object device".
+         * It assigns his unique ID
```
jvermillard: I think this comment have a really low value:
by reading the code we know we are assigning an ID
just remove it :)

jvermillard:
:+1: looks good

PR #22, opened 2015-03-05
obj_acc_ctrl

bojh:
New Object Implementation: Access Control (dummy w/o functionality).
Clean up memory leaks for single objects instanceList.

### DIFF
--- a/tests/client/CMakeLists.txt
+++ b/tests/client/CMakeLists.txt
@@ -22,6 +22,7 @@ SET(SOURCES
     object_location.c
     object_connectivity_moni.c
     object_connectivity_stat.c
+    object_access_control.c
     test_object.c)
 
 add_executable(lwm2mclient ${SOURCES} ${CORE_SOURCES})
--- a/tests/client/lwm2mclient.c
+++ b/tests/client/lwm2mclient.c
@@ -253,10 +253,8 @@ static uint8_t prv_buffer_send(void * sessionH,
     {
         fprintf(stderr, "#> failed sending %lu bytes\r\n", length);
         return COAP_500_INTERNAL_SERVER_ERROR ;
-    } else {
-        conn_s_updateTxStatistic(objArray[7], length, false);
     }
-
+    conn_s_updateTxStatistic(objArray[7], (uint16_t)length, false);
     fprintf(stderr, "#> sent %lu bytes\r\n", length);
     return COAP_NO_ERROR;
 }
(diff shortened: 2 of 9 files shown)


bojh:
There is a open question about the added "Access Control" multiple resource (MR) ACL: If I create an new object instance e.g. with leshan-standalone-server, the ACL-MR has empty instances (leshan is currently only able to set writable resources at create). Reading this resource of the created instance again generates an COAP_INTERNAL_SERVER_ERROR in wakaama core! 
TLV coding for MR'S is only defined with one or more instances (see TS table 19). 
Is this a irregular LWM2M MR status injected by leshan server?
What is the right way to create the response?

dnav:
I did not test this yet but my first though is that as the ACL MR (/2/x/2) is optionnal. So if the ACL MR has empty instance, the reply should not contain this resource.
In the prv_read() for the ACL object, the resList should not include RES_O_ACL if the accCtrlOiP->accCtrlValList is nil.

My 2cts

bojh:
...seems reasonable! 
.. will implement that - but have a short break ;-)

dnav:
Actually, after some thoughts, maybe the ACL mechanism should be included directly in the core. (as a compile-time option) There are a lot of data to be retrieved  when modifying the object and the core will need to access the ACL a lot at each DM operation.
Your thoughts ?

bojh:
Yes there is good reason. My intention was to contribute reasonable PL parts: 1st step: a dummy Access Control CRUD core. A 2nd will be, what you just have proposed.

boaks:
May be a „specialized callback” would be better.

typedef uint16_t (*lwm2m_ get_ac_callback_t)(uint16_t objId, uint16_t instId, uint16_t serverid, void * userData);

and add it a „optional“ parameter to lwm2m_init.
If NULL, the ac is not checked, otherwise the callback is used to fetch the ac value.

Mit freundlichen Grüßen / Best regards

Achim Kraus

Bosch Software Innovations GmbH
Communications (INST/ESY4)
Stuttgarter Straße 130
71332 Waiblingen
GERMANY
www.bosch-si.de
www.blog.bosch-si.com<http://www.blog.bosch-si.com>

Tel. +49 711 811-58139
achim.kraus@bosch-si.com<mailto:achim.kraus@bosch-si.com>

Registered office: Berlin, Register court: Amtsgericht Charlottenburg, HRB 148411 B
Executives: Dr.-Ing. Rainer Kallenbach; Michael Hahn


Von: dnav [mailto:notifications@github.com]
Gesendet: Freitag, 6. März 2015 14:26
An: eclipse/wakaama
Betreff: Re: [wakaama] obj_acc_ctrl (#22)


Actually, after some thoughts, maybe the ACL mechanism should be included directly in the core. (as a compile-time option) There are a lot of data to be retrieved when modifying the object and the core will need to access the ACL a lot at each DM operation.
Your thoughts ?

jvermillard:
@boaks why not, but maybe we want to expunge the whole ACL feature code at compile time for saving some flash size.

boaks:
@jvermillard the compile switch will also work with the callback.
The callback should just separate the processing of the AC and the "data maintaining" of the ACL.
While the processing of the AC should be in the core, I would prefer, if the "data maintaining" is left to the client.

bojh:
OK - I am now able to offer a new version with:
- when the ACL MR is empty a read request would return COAP_404_NOT_FOUND
- a single ACL request ends up in a red "Read" button
- when than a complete object is read, the ACL MR would be skipped.
If it is confirmed that the TS does not specify the behavior - we should ask OMA specify the right way!
Any votes?

dnav:
There are two small memory leaks in connectivity objects close because yoy do not iterate throught the  linked list but it is not a bit deal since these are called at client end of life.

bojh:
OK - I understand your point: To keep it simple - this is the simplest implementation for singleton objects

PR #23, opened 2015-03-11
Add a lwm2m_list_free() utility function.

dnav:


boaks:
I would vote for a comment of the "obvious":

"if the objects in the list needs to do more than just "free()" their instance, don' use "lwm2m_list_free()".

or add an optional callback for additional cleanup.

dnav:
I edited the comment.
Having a callback may be overkill for a simple linked list walk.

PR #24, opened 2015-03-11
Do not a reply to DM requests coming from unregistered servers.

dnav:
This can serve as a work around for issue https://github.com/OpenMobileAlliance/OMA-LwM2M-Public-Review/issues/7
If a request arrives before the registration ACK, it will be ignored. So the server will resend it per CoAP specifications. Chances are the ACK will arrive before the retransmission.

### DIFF
--- a/core/internals.h
+++ b/core/internals.h
@@ -120,7 +120,7 @@ int transaction_handle_response(lwm2m_context_t * contextP, void * fromSessionH,
 coap_status_t handle_dm_request(lwm2m_context_t * contextP, lwm2m_uri_t * uriP, void * fromSessionH, coap_packet_t * message, coap_packet_t * response);
 
 // defined in observe.c
-coap_status_t handle_observe_request(lwm2m_context_t * contextP, lwm2m_uri_t * uriP, void * fromSessionH, coap_packet_t * message, coap_packet_t * response);
+coap_status_t handle_observe_request(lwm2m_context_t * contextP, lwm2m_uri_t * uriP, lwm2m_server_t * serverP, coap_packet_t * message, coap_packet_t * response);
 void cancel_observe(lwm2m_context_t * contextP, uint16_t mid, void * fromSessionH);
 
 // defined in registration.c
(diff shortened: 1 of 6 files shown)


boaks:
This solution has impacts:
- LWM2M servers need to know, that 401 UNAUTHORIZED needs to be handled different in the register phase.
(see also coap 5.9.2.2, 4.01 Unauthorized
"The client is not authorized to perform the requested action. The
client SHOULD NOT repeat the request without first improving its
authentication status to the server.")

- You may get even more traffic during registration

dnav:
> LWM2M servers need to know, that 401 UNAUTHORIZED needs to be handled different in the register phase.
This is the current status. With this change, the LWM2M client does not reply to any request until the registration phase finished. So the server has no special handling to implement.

boaks:
"So the server has no special handling to implement."

Scenario:
- LWM2M client (coap client) registers
- LWM2M server (coap server) response to the register
- LWM2M server (coap client) requests a resource
- LWM2M client (coap server) receives the request for a resource before the response to register (reorder, package loss)
- LWM2M client (coap server) reponse to the resource request with 4.01
- LWM2M server (coap client) receives response with 4.01 

So, with no special handling on the LWM2M server, the LWM2M server may either get 2.05 or 4.01.
.

dnav:
Yes this is the current behavior of wakaama.
With this change you will now have:
* LWM2M client (coap client) registers
* LWM2M server (coap server) response to the register
* LWM2M server (coap client) requests a resource
* LWM2M client (coap server) receives the request for a resource before the response to register (reorder, package loss)
* LWM2M client (coap server) **do not reply to the LWM2M server**
* LWM2M client receives the response to the register
* LWM2M server (coap client) retransmits the request of a resource
* LWM2M client (coap server) receives the request for a resource
* LWM2M client (coap server) replies to the resource request with 2.05 and content
* LWM2M server (coap client) receives response with 2.05 and content

boaks:
OK, your right!

I overseen the **IGNORE**.

(By the way, the "current" behaviour of wakaam was a response with 2.05, because the registration status was only checked for observe-, not for read-requests : - ). )

PR #25, opened 2015-03-12
Wrap around gettimeofday()

dnav:
For easier embedded OSes ports, a lwm2m_gettime() function is provided. On POSIX compliant systems, this a just a wrapper to gettimeofday with a simpler interface.
Also modifies the lwm2m_step() API.

### DIFF
--- a/core/liblwm2m.h
+++ b/core/liblwm2m.h
@@ -511,8 +511,8 @@ lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback, lw
 // close a liblwm2m context.
 void lwm2m_close(lwm2m_context_t * contextP);
 
-// perform any required pending operation and adjust timeoutP to the maximal time interval to wait.
-int lwm2m_step(lwm2m_context_t * contextP, struct timeval * timeoutP);
+// perform any required pending operation and adjust timeoutP to the maximal time interval to wait in seconds.
+int lwm2m_step(lwm2m_context_t * contextP, time_t * timeoutP);
 // dispatch received data to liblwm2m
 void lwm2m_handle_packet(lwm2m_context_t * contextP, uint8_t * buffer, int length, void * fromSessionH);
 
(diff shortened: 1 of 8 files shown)


jvermillard:
+1 timeval was a PITA on MBED

core/liblwm2m.h
```diff
@@ -520,7 +526,7 @@ int lwm2m_configure(lwm2m_context_t * contextP, char * endpointName, char * msis
 int lwm2m_start(lwm2m_context_t * contextP);
 
 // check if the server registrations are outdated and needs to be renewed
-int lwm2m_update_registrations(lwm2m_context_t * contextP, uint32_t currentTime, struct timeval * timeoutP);
+int lwm2m_update_registrations(lwm2m_context_t * contextP, time_t currentTime, time_t * timeoutP);
```
jvermillard: why a time_t pointer here and not the value?

bojh:
+1 for Arduino & Spark ports

core/liblwm2m.h
```diff
@@ -505,8 +511,8 @@ lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback, lw
 // close a liblwm2m context.
 void lwm2m_close(lwm2m_context_t * contextP);
 
-// perform any required pending operation and adjust timeoutP to the maximal time interval to wait.
-int lwm2m_step(lwm2m_context_t * contextP, struct timeval * timeoutP);
+// perform any required pending operation and adjust timeoutP to the maximal time interval to wait in seconds.
+int lwm2m_step(lwm2m_context_t * contextP, time_t * timeoutP);
```
jvermillard: Same: why a time_t pointer here and not the value?

dnav: Because lwm2m_step modify the timeout. (timeout is an OUT parameter).

core/liblwm2m.h
dnav: Actually, this function should update the timeout if a registration is needed. I will change this

eclipsewebmaster:
:white_check_mark: 
#### Eclipse validation results:
All committers passed Eclipse CLA and Signed-off-by validation.

dnav:
Updated.
I also removed lwm2m_update_registration**s**() from the exposed APIs. I could not figure out why it was here.

PR #26, opened 2015-03-16
Fix Bug 447601: Integers in TLV are encoded on 1, 2, 4 or 8 bytes.

dnav:

### DIFF
--- a/core/tlv.c
+++ b/core/tlv.c
@@ -116,6 +116,55 @@ static int prv_create_header(uint8_t * header,
     return header_len;
 }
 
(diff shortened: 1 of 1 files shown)


dnav:
Bugzilla link: https://bugs.eclipse.org/bugs/show_bug.cgi?id=447601 (Yes there is one for wakaama.)

Difference can be seen when reading object /3/0. The value for resource 13 is now encoded on 8 bytes instead of 5.

dnav:
Updated to include @boaks ' fix from PR #28

PR #30, opened 2015-03-31
Support for alternative path in registration

dnav:
As described in the LWM2M TS §5.3.1:
> By default, the RFC6690 links of Objects are located under the root path as in the examples above. However, devices might be hosting other Resources on an endpoint, and there may be the need to place Objects under an alternative path. This is achieved by including an OMA LWM2M link in addition to the Object links as follows, e.g. to place Objects under the “/lwm2m” path:
</lwm2m>;rt="oma.lwm2m", </lwm2m/1/101>, </lwm2m/1/102>, </lwm2m/2/0>, </lwm2m/2/1>, </lwm2m/2/2>, </lwm2m/3/0>,</lwm2m/4/0>,</lwm2m/5>
The RFC6690 Resource Type parameter (i.e., rt="oma.lwm2m") MAY be used to provide the information that the path in front of the Resource Type parameter is used for the LWM2M enabler.

### DIFF
--- a/core/internals.h
+++ b/core/internals.h
@@ -70,6 +70,10 @@
 
 #define LWM2M_MAX_PACKET_SIZE 198
 
+#define REG_ALT_PATH_LINK         "<%s>;rt=\"oma.lwm2m\","
+#define REG_OBJECT_PATH           "<%s/%hu>,"
+#define REG_OBJECT_INSTANCE_PATH  "<%s/%hu/%hu>,"
+
 #define URI_REGISTRATION_SEGMENT        "rd"
 #define URI_REGISTRATION_SEGMENT_LEN    2
 #define URI_BOOTSTRAP_SEGMENT           "bs"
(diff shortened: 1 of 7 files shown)


dnav:
This is a Work In Progress.
The alternative path is not yet supported in DM requests. And it only supports single segment alt. path.

@jvermillard : change marked REVERT-ME is included only for testing purpose. Leshan (as at leshan.eclipse.org) does not support the alternative path in the registration message.

jvermillard:
@dnav leshan support alternative lwm2m path, we use it a lot in our proprietary client, but maybe it's bugged :)

jvermillard:
so yes, it looks broken on the leshan UI side, the core API works.
I sent a read on /test/3/0/0 and received an empty content.

If you want to use the raw json API on leshan, just register your device on leshan.eclipse.org and then issue a HTTP on http://leshan.eclipse.org/api/clients/testlwm2mclient/3/0/0
it should send a CoAP GET on the testlwm2mclient on /test/3/0/0

dnav:
I just tested without the UI and my latest version works fine.

PR #35, opened 2015-04-13
Simplify lwm2m_uri_t by removing its member "flag".

dnav:
An ID is considered not set when it is equal to LWM2M_MAX_ID (0xFFFF).
Object and Object Instance can not have the MAX_ID 0xFFFF per LWM2M TS. We also add this limitation for Resource IDs.

### DIFF
--- a/core/internals.h
+++ b/core/internals.h
@@ -81,12 +81,13 @@
 #define URI_BOOTSTRAP_SEGMENT           "bs"
 #define URI_BOOTSTRAP_SEGMENT_LEN       2
 
-#define LWM2M_URI_FLAG_DM           (uint8_t)0x00
-#define LWM2M_URI_FLAG_REGISTRATION (uint8_t)0x20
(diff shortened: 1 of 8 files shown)


dnav:
Actually you are right MAX_ID is not a valid value for a resource id. It is just SO more confortable to have this limitation that I could not help myself...

In the TS, MAX_ID is also implicitly reserved for Object IDs by the definition of ObjLnk in Appendix C. (This is just my understanding.)

boaks:
"In the TS, MAX_ID is also implicitly reserved for Object IDs by the definition of ObjLnk in Appendix C. (This is just my understanding.)"

TS:
"An Object Link referencing no Object Instance will contain the concatenation of 2 MAX-ID values (null link)"

In my interpretation, this just means, that 0xFFFFFFFF is null. I don't see, that this also reserves 0xFFFF as special value for OBJECT IDs.

dnav:
@boaks I just created a branch "new_bootstrap" for you.

PR #36, opened 2015-04-14
For review only. JSON handling

dnav:
This is a coding/decoding of SENML format with only the LWM2M required tokens.
The code is probably unmaintanable.

This is use only by the TLV test app.

Do Not Merge ! (for now)


boaks:
@dnav 

Looks "not too bad" :-). Seems to be much smaller, then I would have expected.
As far as I understood, the impact on the "application layer read/write" will be limited to multi resource instances. May be we just need to copy the flags from the outer to the sub TLVs in that case.

jvermillard:
JSON/SenML is optional and a large chunk of code, maybe user should be able to easily disable it at compile time?

core/json.c
```diff
@@ -0,0 +1,711 @@
...
+                break;
+
+            default:
+                return -1;
+            }
+        }
+        break;
+
+        default:
+            return -1;
```
jvermillard: maybe we want debug log here for knowing the source of the parse error?

PR #37, opened 2015-04-15
Add data type info to lwm2m_tlv_t

dnav:
Inspired from Bosch SI changes.
Useful for observation parameters and JSON encoding.


bojh:
Nice to see - forward thinking for conditional observe!

PR #38, opened 2015-04-20
New bootstrap with some fixes

boaks:
Cleanup core from backup/restore stuff.
Extend retry handling.
Change API to handle multiple bootstrap servers with server id 0.


dnav:
This one will require some time to review. I'll just make some quick comments on the first read for today.

core/bootstrap.c
```diff
@@ -0,0 +1,215 @@
+/*******************************************************************************
+ *
+ * Copyright (c) 2013, 2014 Intel Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
```
dnav: This code does not originate from Intel so the Copyright header must be changed.

core/management.c
```diff
@@ -59,11 +60,42 @@ coap_status_t handle_dm_request(lwm2m_context_t * contextP,
...
+        if (NULL != bsServerP)
+        {
+            // server initiated bootstrap?
+            // currently not implemented.
+            return NOT_IMPLEMENTED_5_01;
+        }
+        if ( serverP->status != STATE_REGISTERED &&
+                serverP->status != STATE_REG_UPDATE_PENDING)
+        {
+            return COAP_IGNORE;
```
dnav: Keep it like this for now but IMHO a bootstrap can occur anytime.

core/management.c
```diff
@@ -96,7 +128,7 @@ coap_status_t handle_dm_request(lwm2m_context_t * contextP,
         {
             if (!LWM2M_URI_IS_SET_INSTANCE(uriP))
             {
-                result = object_create(contextP, uriP, message->payload, message->payload_len);
+                result = object_create(contextP, uriP, (char*)message->payload, message->payload_len);
```
dnav: FYI: I plan to make a big cleanup and use only uint8_t * instead of this mix with char *

core/management.c
```diff
@@ -167,6 +199,29 @@ coap_status_t handle_dm_request(lwm2m_context_t * contextP,
...
+    {
+        int i;
+        for (i = 0 ; i < context->numObject ; i++)
+        {
+            // Only security and server objects are deleted upon a DEL /
+            switch (objectList[i]->objID)
+            {
+            case LWM2M_SECURITY_OBJECT_ID:
+            case LWM2M_SERVER_OBJECT_ID:
+                objectList[i]->closeFunc(objectList[i]);
```
dnav: calling closeFunc may leave the object plugins in an undefined state.
I would suggest to make the delete function mandatory and call it on all instance if the client is to support the bootstrap.

core/management.c
boaks: A cleanup would be much appreciated :-).

core/bootstrap.c
boaks: achim.kraus@bosch-si.com

boaks: @prieux 
So, would It be possible to get a proper header from you?
I don't think the header should be from my company :-).

core/liblwm2m.h
```diff
@@ -372,7 +373,8 @@ typedef enum
 typedef struct _lwm2m_server_
 {
     struct _lwm2m_server_ * next;   // matches lwm2m_list_t::next
-    uint16_t          shortID;      // matches lwm2m_list_t::id
+    uint16_t          secObjInstID; // matches lwm2m_list_t::id
```
dnav: Did you check it does not impact any call to lwm2m_list_find() ?
ACL mechanism will look for servers by their shortID.

core/liblwm2m.h
```diff
@@ -497,13 +498,22 @@ typedef struct _lwm2m_observed_
     lwm2m_watcher_t * watcherList;
 } lwm2m_observed_t;
 
+typedef enum {
+    NOT_BOOTSTRAPPED = 0,
+    BOOTSTRAP_REQUESTED,
+    BOOTSTRAP_CLIENT_HOLD_OFF,
+    BOOTSTRAP_PENDING,
+    BOOTSTRAP_FINISHED,
```
dnav: How do you determine that a boostrap is finished ?

core/liblwm2m.h
```diff
@@ -497,13 +498,22 @@ typedef struct _lwm2m_observed_
     lwm2m_watcher_t * watcherList;
 } lwm2m_observed_t;
 
+typedef enum {
+    NOT_BOOTSTRAPPED = 0,
```
dnav: equivalent to an empty lwm2m_context_t::serverList (just thinking out loud here)

core/liblwm2m.h
```diff
@@ -497,13 +498,22 @@ typedef struct _lwm2m_observed_
...
+    BOOTSTRAP_REQUESTED,
+    BOOTSTRAP_CLIENT_HOLD_OFF,
+    BOOTSTRAP_PENDING,
```
dnav: meaning that a first message has been received from a bootstrap server ?

core/liblwm2m.h
```diff
@@ -497,13 +498,22 @@ typedef struct _lwm2m_observed_
...
+    BOOTSTRAP_FINISHED,
+    BOOTSTRAP_FAILED,
+    BOOTSTRAPPED
```
dnav: Equivalent to having a least one LWM2M Server

core/liblwm2m.h
boaks: I checked that and I didn't found a match.
ACL is not related to bootstrap servers (as far as I understand).

core/liblwm2m.h
```diff
@@ -555,6 +567,9 @@ int lwm2m_configure(lwm2m_context_t * contextP, char * endpointName, char * msis
 // create objects for known LWM2M Servers.
 int lwm2m_start(lwm2m_context_t * contextP);
 
+// manage a bootstrap session
+int lwm2m_bootstrap(lwm2m_context_t * contextP);
```
dnav: Bootstrap should be automatic when starting the client with an empty server list.

core/liblwm2m.h
boaks: Just with "no message received for COAP_DEFAULT_MAX_AGE" (it's mostly still pascal's implementation).

core/liblwm2m.h
boaks: It's more "since the last restart", the "in general".

core/liblwm2m.h
boaks: That's the BOOTSTRAP_PENDING.

core/liblwm2m.h
boaks: Also more "succesfull BOOTSTRAPED since last reboot". In my opinion, with a potential sequence of bootstrap servers, this may be also valid, if the bootstrap server was exchanged even without a LWM2M Server.

core/liblwm2m.h
dnav: Yes sure ACL is not related to Bootstrap.
It is just the only thing I can see that would need the server short ID.

core/liblwm2m.h
boaks: Yes, therefore see "registrationer.c / registration_update".

core/bootstrap.c
dnav: @jvermillard @prieux I think it should be from Sierra Wireless.

core/liblwm2m.h
boaks: With the secObjInstId it's very easy to access the server short ID. So even if required, it would be easy.

core/bootstrap.c
jvermillard: agree, do you want me to change it on the new_bootstrap branch?

core/liblwm2m.h
jvermillard: @boaks triggering the bootstrap from a public API should not be allowed, it's triggered only if the server list is empty or all the DM registration failed

but I'm not sure the spec won't move

core/bootstrap.c
dnav: Yes please.

jvermillard: @dnav done!

core/liblwm2m.c
```diff
@@ -123,13 +146,38 @@ void lwm2m_close(lwm2m_context_t * contextP)
         LWM2M_LIST_FREE(targetP->watcherList);
 
         lwm2m_free(targetP);
+        targetP = NULL;
```
dnav: This line is useless isn't it ?

core/liblwm2m.c
```diff
@@ -328,3 +374,32 @@ int lwm2m_step(lwm2m_context_t * contextP,
 
     return 0;
 }
+
+#ifdef LWM2M_CLIENT_MODE
+int lwm2m_start(lwm2m_context_t * contextP)
```
dnav: Reviewing your change, I come to realize I don't like the lwm2m_start() API...
Before its only job was to retrieve servers. Something that should be done by lwm2m_configure() or at the end of a bootstrap.
Now it is used as a reset. The client should do this by closing the lwm2m_context_t and create a new one configured with the same objects. The close callbacks for objects should also be removed. Objects data are the responsability of the client.

Your thoughts ?

dnav:
My main concern with this from a design point-of-view is the bootstrap state. In my view, the bootstrap server can interact with the client at any time. And both bootstrap server and LWM2M server requests should be handled.
Do you have the same view ?

boaks:
The TS seems to be "not to clear" for me at that point.
Though bootstrap- and data-request use the same "COAP" mapping, I see only two ways of differentiate such request: end-point or bootstrap-state.
If the TS clarifies, that all security objects must point to different end-points (really different), then removing the bootstrap-state would be an option. But right now, I would stay with it.

dnav:
An endpoint being both the LWM2M server and the Bootstrap server ? In NoSec mode, we clearly do not have mean to differentiate requests. With DTLS involved it is possible *if* they do not share the same security credentials.
But would it really be an issue ? As you said DM and Bootstrap interfaces use the same CoAP mapping. So such an endpoint (both bootstrap and LWM2M server) would be almighty. But it should not impact the client.

core/liblwm2m.c
boaks: I think, don't change too much.
Otherwise a "feature" pull requests will lasts for ever :-)

dnav:
Would it be possible to enclose the bootstrap feature between a compilation flag ?
I plan to merge this quickly. there are too many PR in parallel...

boaks:
Would it be possible to enclose the bootstrap feature between a compilation flag ?

I check that ...

boaks:
Except of some rudiments in "uri.c", the bootstrap stuff is now depending on the compiler switch LWM2M_BOOTSTRAP.
(Though the pull request contains a lot of commits, I didn't rebase it. I just merged it.)

dnav:
Manually merged with two little fixes:
- tests/TLV was not merged from master
- some files were chmoded to 755.

PR #39, opened 2015-04-21
Add support for Float values.

dnav:
Requires the definition of LWM2M_BIG_ENDIAN or LWM2M_LITTLE_ENDIAN since there are no standard ntoh*() for float and double.
new APIS:
* lwm2m_opaqueToFloat()
* lwm2m_tlv_encode_float()
* lwm2m_tlv_decode_float()
* lwm2m_PlainTextToFloat64()

Client's test object (ID: 1024) has a new resource (ID: 3) of type Float for testing.
There is a new test application named "misc" to test text to float/int functions.


dnav:
Are you refering to the removal of snprintf() ? If so, I've done this because some embedded platforms does not provide all the libc.

boaks:
No, I referring the switch with case 4 and case 8 in tlv.c/lwm2m_opaqueToFloat. 
I'm not sure, if the case 4 is required, because the generic case 8 will do it also (when removing the switch/case).

dnav:
OK.
When size is 4, I need to go through a temporary variable before memcpy to the double. 32-bits floats and 64-bits floats are not as compatible as integers...

boaks:
Uups, sorry, I oversaw that detail.

PR #40, opened 2015-04-23
Fix uninitialized counter.

boaks:
Fix an uninitialized counter.

Signed-off-by: Achim Kraus <achim.kraus@bosch-si.com>

### DIFF
--- a/tests/client/object_access_control.c
+++ b/tests/client/object_access_control.c
@@ -153,6 +153,7 @@ static uint8_t prv_read(uint16_t instanceId, int * numDataP,
         }
     }
 
+    ni = 0;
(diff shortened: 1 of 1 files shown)


tests/client/object_access_control.c
```diff
@@ -153,6 +153,7 @@ static uint8_t prv_read(uint16_t instanceId, int * numDataP,
         }
     }
 
+    ni = 0;
```
jvermillard: why not initialize it a line 125 with the declaration?

boaks: Because "nothing" is intialized at definition.
May be something for the cleanup :-).

PR #42, opened 2015-04-24
Code fix for object Access Control read implementation

bojh:
This PR corrects TLV structure initialization for reading the Access Control object in the case, the ACL multiple instance resource does not contain any instances. The following resources when reading the whole object were lost. By the way, the old implementation lacks of a missing counter init potentially producing an core.


jvermillard:
Should close https://bugs.eclipse.org/bugs/show_bug.cgi?id=465314 ?

dnav:
Indeed.

PR #45, opened 2015-04-29
Work In Progress: Bootstrap server

dnav:
The bare minimum at LWM2M level to make a Bootstrap server.
Code-wise this is very close to DM operations.


dnav:
No longer a work in progress.

PR #48, opened 2015-05-06
Implements a simple bootstrap server

dnav:
Bootstrap Information are taken from a .ini file.
It can only perform Delete operations and Write to /0 and /1.
Features a README.

To test it, launch the wakaama client with parameters:
> lwm2mclient -b -p 5685

@boaks : With the example .ini file, if you change the client name (lwm2mclient -b -p 5685 -n something), you'll see the client does not support Delete operations on something else than "/" URI.


boaks:
So, finally I made it to commit 1fc96c1 :-) 

    @boaks : With the example .ini file, if you change the client name (lwm2mclient -b -p 5685 -n 
     something), you'll see the client does not support Delete operations on something else than "/"
     URI.

The error message is:

    Received status 4.00 (COAP_400_BAD_REQUEST) for URI /0 from endpoint test.

This error is caused by the definition of DELETE in the Data Management Interface, where the OID and IID is required. I guess "delete /0" should delete all instances of that object? We could change the implementation of delete accordingly but currently the TS says 8.2.2:

    "The Write or Delete operation targets to an Object Instance or a Resource."
 
(HM, delete a "Resource" will also not work :-). And I'm not sure, how this should be implemented.)

dnav:
> The error message is:
```
Received status 4.00 (COAP_400_BAD_REQUEST) for URI /0 from endpoint test.
```
>This error is caused by the definition of DELETE in the Data Management Interface, where the OID and IID is required. I guess "delete /0" should delete all instances of that object?

I think so. A bootstrap delete on an object should iterate through all instances of that object and call the delete callback on each of them.
> We could change the implementation of delete accordingly but currently the TS says 8.2.2:
```
"The Write or Delete operation targets to an Object Instance or a Resource."
```

IMHO this is another hole in the LWM2M TS. Either you delete all Objects instances in the client or only specific instance.
But what if the Server wants to delete only LWM2M and Bootstrap accounts ? It needs prior knowledge of the server and security objects instances IDs existing in the Client.
Or maybe we should understand that "delete /" concerns only /0 and /1 objects...

boaks:
IMHO this is another hole in the LWM2M TS.

Seems so. And yes, there are more solutions for that issue. Therefore, until the TS didn't solve it in an other way, I would stay with pascal's solution and interpret "delete /" as delete "/0" an "/1", as the current implemetation already did (even if the TS tells something different "delete all except the bootstrap server account").

dnav:
Now ready for review before merge

boaks:
I'm not sure, if I understand the data modell:
- "server" containing security infos (e.g. psk identity, psk secret key)
- "endpoint" referencing "server"s

But, though each "endpoint" may have it's own psk identity/key pair, the reuse of the "server" may be rare.

dnav:
Indeed. But this is a really simple server for demo and testing purposes.

My main usage is to test various clients without having to provision each one. With the same PSK for all of them.

boaks:
Default config file name: "bootstrap_info.ini"
GIT: "bootstrap_server.ini"

dnav:
Fixed.

BTW, notice that the bootstrap_server.c implements a simple state machine to make sure ACKs and commands are sent in the right order.

PR #49, opened 2015-05-07
Introduce "separate response" transactions.

boaks:
Add processing of transactions with separate response as specified in section 2.2 of RFC 7252.
Add first cunit tests and optional memory tracing.

### DIFF
--- a/core/internals.h
+++ b/core/internals.h
@@ -120,11 +120,12 @@ int prv_getRegisterPayload(lwm2m_context_t * contextP, uint8_t * buffer, size_t 
 int object_getServers(lwm2m_context_t * contextP);
 
 // defined in transaction.c
-lwm2m_transaction_t * transaction_new(coap_method_t method, char * altPath, lwm2m_uri_t * uriP, uint16_t mID, lwm2m_endpoint_type_t peerType, void * peerP);
+lwm2m_transaction_t * transaction_new(coap_message_type_t type, coap_method_t method, char * altPath, lwm2m_uri_t * uriP, uint16_t mID, uint8_t token_len, uint8_t* token, lwm2m_endpoint_type_t peerType, void * peerP);
+
 int transaction_send(lwm2m_context_t * contextP, lwm2m_transaction_t * transacP);
 void transaction_free(lwm2m_transaction_t * transacP);
 void transaction_remove(lwm2m_context_t * contextP, lwm2m_transaction_t * transacP);
-int transaction_handle_response(lwm2m_context_t * contextP, void * fromSessionH, coap_packet_t * message);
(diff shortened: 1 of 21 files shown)


core/internals.h
```diff
@@ -120,11 +120,12 @@ int prv_getRegisterPayload(lwm2m_context_t * contextP, uint8_t * buffer, size_t
 int object_getServers(lwm2m_context_t * contextP);
 
 // defined in transaction.c
-lwm2m_transaction_t * transaction_new(coap_method_t method, char * altPath, lwm2m_uri_t * uriP, uint16_t mID, lwm2m_endpoint_type_t peerType, void * peerP);
+lwm2m_transaction_t * transaction_new(coap_message_type_t type, coap_method_t method, char * altPath, lwm2m_uri_t * uriP, uint16_t mID, uint8_t token_len, uint8_t* token, lwm2m_endpoint_type_t peerType, void * peerP);
```
dnav: Why add the type parameter when a transaction is always a CON message ?

boaks:
If a “transaction” message always uses “CON”, seems for me to be related to the exact definition of the term “transaction”.
Using “transaction” only for “RFC 7252, section 2.1. Messaging Model”, the extension to a “separate response” would be even simply wrong. 
Using “transaction” for “RFC 7252, section 2.2.  Request/Response Model”, the “NON” case is mentioned on page 14 as the last pattern of this section.
Though Wakaama uses the transactions for the request/response model (e.g. registration), I extended the processing towards the RFC.

dnav:
There are two sides on this question:
First we need to accept all kinds of request/response a peer could use: piggyback, separate and NON.
Then we need to decide what we will use. For now we always send requests with CON and we always reply with piggyback.
With this change, how do the core decide to use a CON or a NON ? (Sending NON is just a way to force the peer to not send separate response.)

boaks:
You are right,(currently and this may be never changing) the core needs to decide, what to use, and it decides to uses CON. Also the LWM2M TS limits the MUST support of NON to the observe/notify interface. I just decided to include it to compelete the feature according to the RFC. 
But if you preferre, I can als remove the "coap_message_type_t" from "transaction_new" (but don't expect too many resources to be freed ;-).

dnav:
So let's keep the message type parameter.

dnav:
I just manually merged this PR in master.
There was a conflict in packet.c. Let me know if I break anything.

boaks:
I checked it. 
The call to "transaction_handle_response" was removed for RST messages. Therefore the transaction is not removed, if a client answers with RST. But I have to confess, that the current transactions callback handler also requires some more work :_). I hope, I can prepare a fix this afternoon.

PR #50, opened 2015-06-02
Support for IPv6 and IPv4.

dnav:
Also remove packet size limitation.


boaks:
Your right, who need something "special", may use overwriting "lwm2m_malloc".

By the way: any plans for merging the current PR?

dnav:
> By the way: any plans for merging the current PRs?

Yes: find time...
I'll comment yours.
Regarding mine: The bootstrap server one requires more work. The JSON one will not be merged as-is. And I keep the lwm2m_uri_t one in case the TS evolves on the reserved IDs.

PR #59, opened 2015-10-08
Move memory trace implementation from the core to tests/utils.

dnav:


dnav:
Indeed.
Allow me to rethink this.

PR #61, opened 2015-11-26
Add a secure client sample

sbernard31:
We add a new secure client example using **tinydtls**. It is largely inspired by the **lightclient** example.
Currently, it support only **PSK** and must be **bootstrapped**. 

Next feature could be to make the bootstrap step optionnal and support RPK too. (maybe also having a way to persist security and server object)

Signed-off-by: Simon Bernard <sbernard@sierrawireless.com>
Also-by: Julien Vermillard <jvermillard@sierrawireless.com>


dnav:
Thanks Simon but I'm pretty sure we can not include the tinydtls source code in the wakaama repo. I have to check with the Eclipse legal.
BTW, you can also find a DTLS enabled client in the IPSO branch (the tests/client with compilation flag WITH_TINYDTLS. To build it, you need tinydtls to be installed in your sysroot). It is less elegant than yours as the DTLS handling is not split in a separate file. I am waiting for tinydtls to be part of the Eclipse Foundation before pushing anything to master.

jvermillard:
Since we have the agreement for "works-with" we can integrate example depending on tinydtls.
But yes we should not commit the code in the repo and let the user download it.

jvermillard:
@dnav I prefer our implementation :) we fixed a bunch of issues (IPv6 with IPv4 address compare) and made the security and server object writable for bootstrap.

boaks:
Is it possible to get some information on compatibility checks?

Does it work with "DTLS role change" 
(discussed in the californium/scandium e-mail dev-list, see
http://dev.eclipse.org/mhonarc/lists/cf-dev/msg00436.html)?

What happens, if one site restarts without inform the other side (crashs,
Does this rely on exchanging "DTLS Alerts (e.g. CLOSE_NOTIFY)")?

jvermillard:
if it's the device which crash after hand shake restarting with a fresh Client Hello will work with scandium as a server. (if the device crash during the handshake is another story, simon is trying to identifying a possible scandium bug here).

I reported some tinydtls bugs on packet loss:
https://sourceforge.net/p/tinydtls/tickets/42/
https://sourceforge.net/p/tinydtls/tickets/43/

Also I didn't tested the case where you restart the server (but everybody knows servers never crash :)
So we still have some work on this front to get it bullet-proof.

dnav:
@sbernard31, could you make a commit without the tinydtls source code ? Thanks.

sbernard31:
I remove the tinydtls source code and add some option to use secure client without bootstrap.
(before integrated this in master, we should probably squash this in 1 commit, I don't do it to make the review easier)

tests/client/object_security.c
```diff
@@ -48,7 +48,7 @@
 
 #define LWM2M_SECURITY_URI_ID                 0
 #define LWM2M_SECURITY_BOOTSTRAP_ID           1
-#define LWM2M_SECURITY_SECURITY_ID            2
+#define LWM2M_SECURITY_MODE_ID            2
```
dnav: These constants are actually defined in multiple places. My bad...
This change should be made in liblwm2m.h. client, lightclient and bootstrap server codes would need to be updated.

tests/secureclient/object_security.c
```diff
@@ -0,0 +1,505 @@
...
+        return COAP_205_CONTENT;
+
+    case LWM2M_SECURITY_SMS_SECRET_KEY_ID:
+        dataP->value = (uint8_t*)targetP->smsSecret;
+        dataP->length = targetP->smsSecretLen;
+        dataP->flags = LWM2M_TLV_FLAG_STATIC_DATA;
+        dataP->dataType = LWM2M_TYPE_OPAQUE;
+        return COAP_205_CONTENT;
+
+    // HOLYSHIT THIS IS SO BROKEN OMA :(
```
dnav: Hum...

jvermillard: Fixed :D

PR #64, opened 2015-12-08
Remove Close callback from lwm2m_object_t.

dnav:
As the client application is creating the objects, the core should not be in charge of freeing them.

### DIFF
--- a/tests/client/lwm2mclient.h
+++ b/tests/client/lwm2mclient.h
@@ -35,28 +35,33 @@ extern int g_reboot;
  * object_device.c
  */
 extern lwm2m_object_t * get_object_device(void);
+void free_object_device(lwm2m_object_t * objectP);
 uint8_t device_change(lwm2m_data_t * dataArray, lwm2m_object_t * objectP);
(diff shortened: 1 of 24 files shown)


tests/client/lwm2mclient.h
```diff
@@ -64,12 +69,14 @@ extern void copy_server_object(lwm2m_object_t * objectDest, lwm2m_object_t * obj
...
  */
 extern lwm2m_object_t * get_object_conn_m(void);
+void free_object_conn_m(lwm2m_object_t * objectP);
 uint8_t connectivity_moni_change(lwm2m_data_t * dataArray, lwm2m_object_t * objectP);
 
 /*
  * object_connectivity_stat.c
  */
 extern lwm2m_object_t * get_object_conn_s(void);
+void free_object_conn_s(lwm2m_object_t * objectP);
```
jvermillard: extern?

dnav: not needed.
The extern are redundant here. I could have removed them.

PR #67, opened 2015-12-15
Rework of the client start sequence

dnav:
When starting in Client mode, there is now a state machine. If LWM2M Servers are provisionned, the core starts to register to them. If all the registrations fail or if there are no LWM2M Servers, we start the bootstrap process. At the end of the bootstrap we return to the initial state.
The bootstrap state is now stored per server. This allows multiple Bootstrap Servers. The bsTimer is gone (for now).
Regarding Bootstrap Operations, the "Delete /" now deletes all instances of all objects. "Delete /n" is now supported.

This is still a Work In Progress for review until next year.
TODOs:
- Unsecure Bootstrap servers accounts cleanup
- Decide client behavior regarding registered LWM2M Servers after a Server Initiated Bootstrap.
- Client can enter a infinite loop if the registration fails
- Fix secure client.
- Test concurrency with multiple Bootstrap Servers.

### DIFF
--- a/core/bootstrap.c
+++ b/core/bootstrap.c
@@ -253,10 +253,6 @@ static coap_status_t prv_check_server_status(lwm2m_server_t * serverP)
 {
     switch (serverP->status)
     {
-    case STATE_DEREGISTERED:
-        // Should not happen
-        return COAP_IGNORE;
-
     case STATE_BS_HOLD_OFF:
         serverP->status = STATE_BS_PENDING;
         break;
@@ -266,6 +262,8 @@ static coap_status_t prv_check_server_status(lwm2m_server_t * serverP)
         serverP->status = STATE_BS_PENDING;
         break;
 
+    case STATE_DEREGISTERED:
+        // server initiated bootstrap
     case STATE_BS_PENDING:
         // do nothing
         break;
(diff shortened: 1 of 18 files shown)


jvermillard:
I think the spec now specify we can have "at most" 1 bootstrap server

jvermillard:
Checked draft of 11 December: "The LWM2M Client MUST have at most one LWM2M Bootstrap Server Account."

dnav:
You're right. I missed that.
So how can you do a device handover ? The previous owner bootstrap server can not provision the new owner bootstrap server information with this restriction.

dnav:
OK, here is my understanding:
In Bootstrap Interface, "Delete /" deletes everything but the Bootstrap Server Account. "Delete /0" deletes everything including the the Bootstrap Server Account.
This allows a Secure Bootstrap Server needs to provision another Secure Bootstrap Server Account.

dnav:
Adding support for Bootstrap when in STATE_RUNNING.
During the bootstrap, existing servers which are deleted or modified are tagged as dirty. During the refresh_server(), we anly create new lwm2m_server_t instances for newly provisionned servers.

PR #76, opened 2016-02-08
Observe conditions

dnav:
On the server-side, a new API to set or clear link attributes.
On the client-side, these attributes are stored in the core as parameters to the observation (their only use for now).

To test: launch the tests/lwm2mclient with the -c command-line argument.
On the server, observe the resource /3/0/9 and use the new "time" command to set pmin and pmax and "attr" command to set lt, gt and stp.


dnav:
Ready for merge.

Missing bit is inherited conditions. But this part is not clear in the TS.

PR #82, opened 2016-02-18
Add IPv4 mode for lwm2mclient

crenz:
corresponding command-line parameter -4 to lwm2mclient.c

The default is to establish an IPv6 connection, so the behavior without
the new parameter is the same as in previous versions.

This change is one of several upcoming changes that I implemented to run wakaama on iOS devices.


crenz:
Sorry, did not sign off correctly. Will do another pull request.

PR #83, opened 2016-02-18
Add IPv4 mode for lwm2mclient

crenz:
corresponding command-line parameter -4 to lwm2mclient.c

The default is to establish an IPv6 connection, so the behavior without
the new parameter is the same as in previous versions.

Signed-off-by: Christian Renz <crenz@web42.com>


dnav:
Hi,
This is an useful change, thanks.
I would second @vtrushkin. I'm not really found of global variables...

jvermillard:
This should be done for secureclient also?

crenz:
vtrushkin, makes sense. I will adapt the code accordingly and make another pull request.

Issue #84, opened 2016-02-19
Instance ID is no longer required in Bootstrap Write operation.


dnav:
The client needs to handle this in handle_bootstrap_command() case COAP_PUT.

dnav:
Fixed by #97

PR #85, opened 2016-02-21
Add IPv4 mode for test servers & clients

crenz:
corresponding command-line parameter -4 to lwm2mclient.c, as outlined in previous pull request #83 .

I updated the implementation to pass ai_family as parameter instead of using a global. Also, I ported the changes to dtlsconnection.c/.h, and added the same command-line option to the server, lightclient, secure client and bootstrap server.

*Note*: There is one issue I encountered: It is not possible to use "localhost" as server address in IPv4 mode on my system. Instead, "127.0.0.1" needs to be specified explicitly.

The underlying root cause is using AF_UNSPEC in the hints structure when looking up the server address in connection_create(). getaddrinfo on my OS X system will return [::1] as first IP address for "localhost" in this case. The way to fix this would be to use either AF_INET or AF_INET6 depending on the mode of the client, but I'd still need to investigate a bit to find an elegant way to bring this information into connection_create().

### DIFF
--- a/core/bootstrap.c
+++ b/core/bootstrap.c
@@ -94,7 +94,7 @@ static void bootstrap_initiating_request(lwm2m_context_t * context,
 
     if (bootstrapServer->sessionH == NULL)
     {
-        bootstrapServer->sessionH = context->connectCallback(bootstrapServer->secObjInstID, context->userData);
+        bootstrapServer->sessionH = lwm2m_connect_server(bootstrapServer->secObjInstID, context->userData);
     }
 
     if (bootstrapServer->sessionH != NULL)
@@ -212,7 +212,7 @@ void bootstrap_start(lwm2m_context_t * contextP)
         targetP->status = STATE_DEREGISTERED;
         if (targetP->sessionH == NULL)
         {
-            targetP->sessionH = contextP->connectCallback(targetP->secObjInstID, contextP->userData);
+            targetP->sessionH = lwm2m_connect_server(targetP->secObjInstID, contextP->userData);
         }
         targetP = targetP->next;
     }
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -57,26 +57,14 @@
 #include <stdio.h>
 
 
-lwm2m_context_t * lwm2m_init(lwm2m_connect_server_callback_t connectCallback,
-        lwm2m_buffer_send_callback_t bufferSendCallback,
-        void * userData)
+lwm2m_context_t * lwm2m_init(void * userData)
 {
     lwm2m_context_t * contextP;
-
-    if (NULL == bufferSendCallback)
-        return NULL;
-
-#ifdef LWM2M_CLIENT_MODE
-    if (NULL == connectCallback)
-        return NULL;
-#endif
 
     contextP = (lwm2m_context_t *)lwm2m_malloc(sizeof(lwm2m_context_t));
     if (NULL != contextP)
     {
         memset(contextP, 0, sizeof(lwm2m_context_t));
-        contextP->connectCallback = connectCallback;
-        contextP->bufferSendCallback = bufferSendCallback;
         contextP->userData = userData;
         srand(time(NULL));
         contextP->nextMID = rand();
--- a/core/liblwm2m.h
+++ b/core/liblwm2m.h
@@ -107,6 +107,23 @@ int lwm2m_snprintf(char * str, size_t size, const char * format, ...);
 void lwm2m_printf(const char * format, ...);
 #endif
 
+// communication layer
+#ifdef LWM2M_CLIENT_MODE
+// Returns a session handle that MUST uniquely identify a peer.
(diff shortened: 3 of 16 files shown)


boaks:
Note: There is one issue I encountered: It is not possible to use "localhost" as server address in 
    IPv4 mode on my system. Instead, "127.0.0.1" needs to be specified explicitly.

So can you please give us some overview about your passed (and failed) test scenarios including the used OS for the tests?

boaks:
May be adding the address familiy also to 

    connection_create()

will help to solve that issue. 
If so, please obey, that the function is called from a "callback". Usually a callback should access additional data via a "context", which is passed in using lwm2m_init.
So extending e.g. "client_data_t" with the address familiy would be a good way (from my view).

crenz:
So can you please give us some overview about your passed (and failed) test scenarios including the used OS for the tests?

The OS is OS X 10.11.3, with standard network settings (i.e., IPv6 enabled) and an IPv4 connection to the outside world. Here are my results:


Current master, local lwm2mserver:
    ./lwm2mclient                          # Successfully connects to local server (as expected)
    ./lwm2mclient -h localhost             # Successfully connects to local server (as expected)
    ./lwm2mclient -h ::1                   # Successfully connects to local server (as expected)
    ./lwm2mclient -h 127.0.0.1             # Fails to connect to local server (as expected)
    ./lwm2mclient -h leshan.eclipse.org    # Fails to connect to leshan server (as expected, no IPv6 connection)

Pull request #83, local lwm2mserver in IPv6 mode:
    ./lwm2mclient                          # Successfully connects to local server (as expected)
    ./lwm2mclient -h localhost             # Successfully connects to local server (as expected)
    ./lwm2mclient -h ::1                   # Successfully connects to local server (as expected)
    ./lwm2mclient -h 127.0.0.1             # Fails to connect to local server (as expected)
    ./lwm2mclient -h leshan.eclipse.org    # Fails to connect to leshan server (as expected, no IPv6 connection)
# I.e. - test results are exactly the same as for current master

Pull request #83, local lwm2mserver in new IPv4 mode (lwm2mserver -4):
    ./lwm2mclient -4                       # Successfully connects to local server (as expected)
    ./lwm2mclient -4 -h localhost          # Fails to connect to local server *this is the case I am talking about above*
    ./lwm2mclient -4 -h ::1                # Fails to connect to local server (as expected)
    ./lwm2mclient -4 -h 127.0.0.1          # Successfully connects to local server (as expected)
    ./lwm2mclient -4 -h leshan.eclipse.org # Successfully connects to leshan server (as expected)

boaks:
@crenz 
Thank's for the tests! I hope this helps a lot of user and developer to use it proper.

crenz:
It looks much better now. But it looks like a Java code =) What do you think about following snippet?

Thanks for your suggestion, I updated the code.

dogonthesun:
Thank you. Could you rebase your branch and squash all commits into one?

crenz:
Rebased & squashed. Still not familiar with all git intricacies, hopefully I did the right thing (I followed these instructions: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request )

dogonthesun:
Everything is cool. I just want to note about ai_family. Look around. There are a lot of variables, whose names are corresponded to camelCase convention.

crenz:
Everything is cool. I just want to note about ai_family. Look around. There are a lot of variables, whose names are corresponded to camelCase convention.

That's true, however, the meaning of the variable is not something newly invented within wakaama, but taken directly from the addrinfo struct - where the name is given as "ai_family". So IMHO it makes sense to use exactly the same name to convey that the meaning is exactly the same.

boaks:
From crenz:

> Note: There is one issue I encountered: It is not possible to use "localhost" as server address in IPv4 mode on my system. Instead, "127.0.0.1" needs to be specified explicitly.
The underlying root cause is using AF_UNSPEC in the hints structure when looking up the server address in connection_create(). getaddrinfo on my OS X system will return [::1] as first IP address for "localhost" in this case. The way to fix this would be to use either AF_INET or AF_INET6 depending on the mode of the client, but I'd still need to investigate a bit to find an elegant way to bring this information into connection_create().

I'm still afraid, that this is not only the case for "localhost", it may be the case for all named host.
So this PR will only work with numeric address (e.g. xxx.yyy.nnn.mmm), when ipv4 is switched on.

And modifying "connection_create()" may fix this. How the parametes are passed in is described (
If so, please obey, that the function is called from a "callback". Usually a callback should access additional data via a "context", which is passed in using lwm2m_init.
So extending e.g. "client_data_t" with the address familiy would be a good way (from my view).)

crenz:
boaks, you are right. This issue will appear with any host that has both IP6 and IP4 DNS entries.

I have implemented your suggested addition, and the clients seem to work fine on my machine now.

crenz:
Hmm. Something went wrong during rebasing, it seems. The change for pull request #84 got tied up in this commit as well.

PR #88, opened 2016-02-22
Fix a logic error in utils_findBootstrapServer().

ieei0214:
Fix a logic error in utils_findBootstrapServer().


boaks:
Thanks for the fix!

PR #89, opened 2016-02-24
Fix for issue #84: Instance ID is no longer required in Bootstrap Wri…

ieei0214:
…te operation.

Following a specification change, the instance Id is not required.

If no Resource ID is indicated, then the value included payload is an Object Instance containing the Resource values.

Signed-off-by: Ricky Liu <ieei.liu@gmail.com>

### DIFF
--- a/core/objects.c
+++ b/core/objects.c
@@ -339,6 +339,7 @@ coap_status_t object_create(lwm2m_context_t * contextP,
     lwm2m_data_t * dataP = NULL;
     int size = 0;
     uint8_t result;
+    int i;
 
     if (length == 0 || buffer == 0)
     {
@@ -365,7 +366,18 @@ coap_status_t object_create(lwm2m_context_t * contextP,
 
     size = lwm2m_data_parse(buffer, length, format, &dataP);
     if (size == 0) return COAP_500_INTERNAL_SERVER_ERROR;
-    result = targetP->createFunc(uriP->instanceId, size, dataP, targetP);
+
+    if( dataP->type == LWM2M_TYPE_OBJECT_INSTANCE)
+    {
+        for (i = 0 ; i < size ; i++)
+        {
+            result = targetP->createFunc(uriP->instanceId, dataP->length, (lwm2m_data_t *)(dataP[i].value), targetP);
+        }
+    }
+    else
+    {
+        result = targetP->createFunc(uriP->instanceId, size, dataP, targetP);
+    }
     lwm2m_data_free(size, dataP);
 
     return result;
--- a/core/bootstrap.c
+++ b/core/bootstrap.c
@@ -341,7 +341,15 @@ coap_status_t handle_bootstrap_command(lwm2m_context_t * contextP,
     {
     case COAP_PUT:
         {
-            if (LWM2M_URI_IS_SET_INSTANCE(uriP))
+            if (!LWM2M_URI_IS_SET_INSTANCE(uriP))
+            {                
+                result = object_create(contextP, uriP, format, message->payload, message->payload_len);
+                if (COAP_201_CREATED == result)
+                {
+                    result = COAP_204_CHANGED;
+                }            
+            }
+            else
             {
                 if (object_isInstanceNew(contextP, uriP->objectId, uriP->instanceId))
                 {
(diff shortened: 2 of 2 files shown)


core/objects.c
```diff
@@ -365,7 +366,18 @@ coap_status_t object_create(lwm2m_context_t * contextP,
 
     size = lwm2m_data_parse(buffer, length, format, &dataP);
     if (size == 0) return COAP_500_INTERNAL_SERVER_ERROR;
-    result = targetP->createFunc(uriP->instanceId, size, dataP, targetP);
+
+    if( dataP->type == LWM2M_TYPE_OBJECT_INSTANCE)
+    {
+        for (i = 0 ; i < size ; i++)
+        {
+            result = targetP->createFunc(uriP->instanceId, dataP->length, (lwm2m_data_t *)(dataP[i].value), targetP);
```
dnav: You should use dataP->id instead of uriP->instanceId.

dnav:
Thanks for your input.
In objects.c, it would be better to check if the instance ID is specified either in the URI or in the payload. Then you can check if it already exists or, if it is not specified, determine a new one. 

Regards,

ieei0214:
Sorry I miss this condition if  instance ID  is specified in the payload.

I'll fix it.

ieei0214:
Hi,

When I try to modify the instance 0(Security and server), I got one problem.

When lw_step from "STATE_BS_FINISHED" changes to
"STATE_INITIAL", refresh_server_list() can't find serverList and then do
"STATE_BOOTSTRAP_REQUIRED" again.

Here is the situation.

[image: 內置圖片 1]

Step 1:  Client send /bs to BS server.
Step 2: Server send /0 instance 0
Step 3: Server send /1 instance 0
Step 4: /bs finish bootstrap

After I modify the instance 0 security,  I tag the modified server  into
"STATE_DIRTY".

But after step4, the bs server will be assign  "STATE_BS_FINISHED".

When lw_step from "STATE_BS_FINISHED" to "STATE_INITIAL", the state machine
will trigger refresh_server_list().

But refresh_server_list() will only delete "STATE_DIRTY" server.

That means it won't go into this case in object_getServers line 680.

[image: 內置圖片 2]

Any idea or I miss something?

2016-03-01 21:27 GMT+08:00 dnav <notifications@github.com>:

> Thanks for your input.
> In objects.c, it would be better to check if the instance ID is specified
> either in the URI or in the payload. Then you can check if it already
> exists or, if it is not specified, determine a new one.
>
> Regards,
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eclipse/wakaama/pull/89#issuecomment-190724228>.
>



-- 
RickyLiu

dnav:
Sorry your images are missing.
If I get you right, the Bootstrap Server writes a LWM2M Server at /0/0 and /1/0, overwriting the Bootstrap Server account. This is a tricky thing to do...
I assume you tag the server as dirty in handle_bootstrap_command() case COAP_PUT.

I'm surprised you get this result as step 4 should assign "STATE_BS_FINISHED" only if previous state was "STATE_BS_PENDING".

ieei0214:
> If I get you right, the Bootstrap Server writes a LWM2M Server at /0/0 and /1/0, overwriting the Bootstrap Server account. This is a tricky thing to do...

Yes.  The bs server send /0 and /1 without instance id. But after parse TLV, dataP->id is 0.
Leshan server will send "delete all" after /bs.  But the server that I connect to doesn't send "delete all".
It's strange.

> I assume you tag the server as dirty in handle_bootstrap_command() case COAP_PUT.

Yes. After I modify instance 0 I put diurty on server.

> I'm surprised you get this result as step 4 should assign "STATE_BS_FINISHED" only if previous state was "STATE_BS_PENDING".

I think because the BS registration is done in step 4.  I'll show you the images in wakaama-dev mail group.

dnav:
I reproduced the problem.
This issue is unrelated to your changes. The STATE_DIRTY is lost in handle_bootstrap_command() line 408.
Please go on with your changes. I'm entering an issue (and thinking of a fix) for this particular problem.

Regards

ieei0214:
open new pull request. close here.

PR #114, opened 2016-03-12
Reorganize

davidgraeff:
I propose these changes to the directory structure and file nameing:

* Rename README to README.md: Use the markdown extension to allow github to render the readme file and to be consistent with contributing.md
* Rename tests/ directory to examples/, because there are mostly demonstrational/example projects inside your current test directory. At least for me this caused confusion, because you do not provide documentation and the first thing I was looking for was an example.
* Move the unittests directory now located in examples to "tests/".


dnav:
Hi,

It would be nice indeed to split examples from tests. Some remarks on your changes:
- first as I just merged to master some pending change requests, you''ll need to resolve some conflicts.
- I would prefer to keep misc, TLV and unittests under tests and move only bootstrap_server, client, lightclient, secureclient, server and utils to the example folder.
- be careful that utils/tinydtls is a git submodule. I never tried to mode one and there may be some subtleties.

Regards

davidgraeff:
I'll prepare a new pull request

PR #116, opened 2016-03-17
Reorganize v2

davidgraeff:
I propose these changes to the directory structure and file nameing:
* Rename README to README.md: Use the markdown extension to allow github to render the readme file and to be consistent with contributing.md
* Move example applications out of tests/ into examples/.
* Move examples/utils/connection* to platforms/Linux/

I propose to make full test suites out of examples/TLV and examples/misc:
* Convert examples/TLV /decode.c to a test case.
* Convert examples/misc/test.c to a test case.

I propose to change the build system slightly: 
* Core library and platform part are own cmake "targets" that are linked to the examples. See 0f4d472 for information.

Bugs and warnings:
* Fix a bug in core/tlv.c (a66f500)
* Fix some warnings about constness and missing switch cases.

Because of the discovery branch merge request not all commits will apply cleanly anymore, but I nevertheless would like you to look over the commits if there is something useful for you.


dnav:
Thank you.
I took the liberty to split your pull request in #117 for the fixes and #118 for the directory move.
It is very nice what you did with the unit tests.
I am not convinced by the build system refactoring. It is very powerful and useful but I find it to complex for an example build.
I wait for others opinion on this.

Regards

davidgraeff:
### Buildsystem
In this state it is a bit more complex indeed. I wanted to preserve the current way of using it. 
The complexity of the CMakeLists.txt files, especially in the examples and platform directory, can be reduced if you make only the top level and the library directory a valid target for cmake.
I may show you an alternative to this implementation.

But i very much like the idea of building the library as an independant unit (say: static library) for example to have more strict compiler definitions, other optimisations, detect broken compiler revisions and prohibit those and so on. Because of the way you implement the server and client part of lwm2m (as unified source files), the complexity of the library CMakeLists.txt cannot be reduced, I guess. The automatic endianess detecion can be removed of course.

### Tests
* The decode.c file was a tricky one. The test suite that I made out of it passes for all tests, after i made one little change on how objects are threated in json.c and before the json improvements branch was merged. In the current state, some tests fail. I will look into it again.

* Because I got no response in the mailing list: Is it possible to use gtest instead of cunit? CUnit needs a lot of boilterplate code.

* I would like to have a CI like travis enabled, so that broken merge requests can easily be detected.

Thanks

dnav:
The thing with building the core as a library is that you need several versions of it depending on the options you activate. But please keep in mind that the examples are just examples not deliveries. If someone wants to include wakaama in a Linux distro (for instance), he would have to take care of the packaging.

Switching to gtest requires a validation from the Eclipse Foundation. This can take several weeks. I'm not familiar with gtest, what are the benefits ?

For the CI, let me check what is available in the Eclipse infra.

jvermillard:
For gtest, it's a C++ framework, you can test C but you need to write C++ : https://meekrosoft.wordpress.com/2009/11/09/unit-testing-c-code-with-the-googletest-framework/

I'm not sure to like the idea of writing C++ :)

dnav:
Says the guy who write Java for a living ;)

jvermillard:
I write java for not wirtting C++ :D

davidgraeff:
Guys c'mon, there are advantages in c++ and in java :)
GTest is C++ that is correct, so a c++ compiler have to be used to compile the tests. You can continue to write tests in C.

Regarding the advantages, look at the example here: https://github.com/google/googletest/blob/master/googletest/samples/sample1_unittest.cc.
* Compared to CUnit no boilerplate code (table of test functions, manually registering those in a main() etc) is necessary.
* Writing a new test case is as easy as adding a file new_test_suite.cpp and adding your test
```
TEST(TestSuiteName, TestName) {
  int a = test_important_func();
  EXPECT_EQ(1, a);
}
```
This of course only works seamlessly, if cmake collects all cpp files in tests/ and add it to the test binary.

davidgraeff:
> The thing with building the core as a library is that you need several versions of it depending on the options you activate. But please keep in mind that the examples are just examples not deliveries. If someone wants to include wakaama in a Linux distro (for instance), he would have to take care of the packaging.

You need some variants indeed, this is because the code base unites server, client and bootstrap functionality.
* Compile definitions like LWM2M_BIG_ENDIAN and LWM2M_LITTLE_ENDIAN can be determined automatically or provided by a cross compile toolchain file.
* Compile defintions which are adding features to the client variant are for example: LWM2M_BOOTSTRAP and LWM2M_SUPPORT_JSON. I've implemented cmake targets for all relevant combinations. But such a target can be altered in the applications CMakeLists.txt with a line like this: ``target_compile_definitions(libLwm2m_client PUBLIC -DLWM2M_SUPPORT_JSON)``. So there is no need to have a variant for each possible combination of current or future flags.

But I do understand the concerns. I'm fine if the consens is to not accept this change.

davidgraeff:
Exctracted the idea of using GTest in issue #125, automatically determine endianess in #126, don't repeat source lists in all example and test projects in #127.

PR #133, opened 2016-03-31
Make examples win32 compatible

davidgraeff:
This pull request contains only minimal invasive changes to make
the platform sources win32 compatible. The third commit provides
visual studio compatibility: The VS compiler does not provide unistd.h with getopt.

The entire pull request allows the light client example to be build
on windows with visual studio.

Benefits:
- Windows users can setup a working wakaama application and
  may contribute to the core library more easily.
- The enhanced debugging tools of Visual Studio allow for example to analyse the
  memory profile of the core out of the box. 

Please note: Windows does not allow stdin for select() so all examples with an interactive
shell cannot be build for windows without more invasive changes.

If accepted: It may make sense in that case to rename platforms/Linux to platforms/Posix or something similar. The socket api is posix and platform calls, too.

### DIFF
--- a/platforms/Linux/connection.c
+++ b/platforms/Linux/connection.c
@@ -21,11 +21,22 @@
 #include <ctype.h>
 #include "connection.h"
 
+#ifdef _WIN32
+#define close(s) closesocket(s)
+#endif
+
 // from commandline.c
 void output_buffer(FILE * stream, uint8_t * buffer, int length, int indent);
 
 int create_socket(const char * portStr, int addressFamily)
 {
+    #ifdef _WIN32
+    struct WSAData d;
+    if (WSAStartup(MAKEWORD(2, 2), &d) != 0) {
+	    return -1;
+    }
+    #endif
+
     int s = -1;
     struct addrinfo hints;
     struct addrinfo *res;
@@ -140,7 +151,7 @@ connection_t * connection_create(connection_t * connList,
         close(s);
     }
     if (NULL != servinfo) {
-        free(servinfo);
+        freeaddrinfo(servinfo);
     }
 
     return connP;
--- a/README.md
+++ b/README.md
@@ -1,10 +1,93 @@
-Wakaama (formerly liblwm2m) is an implementation of the Open Mobile Alliance's LightWeight M2M
-protocol (LWM2M).
+Wakaama (formerly liblwm2m) is an implementation of the OMA LightWeight M2M
+protocol (LWM2M) in C suitable for gateway devices as well as embedded devices.
+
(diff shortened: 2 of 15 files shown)


platforms/Linux/connection.c
```diff
@@ -36,7 +47,8 @@ int create_socket(const char * portStr, int addressFamily)
     hints.ai_socktype = SOCK_DGRAM;
     hints.ai_flags = AI_PASSIVE;
 
-    if (0 != getaddrinfo(NULL, portStr, &hints, &res))
+	int r = getaddrinfo(NULL, portStr, &hints, &res);
+    if (0 != r)
```
dnav: Is this useful ?

dnav:
Nice one.
What about adding a small README to explain how to build for Visual Studio ?

platforms/Linux/connection.c
davidgraeff: No not at all. Had to do that while debugging to get the error code. I'll remove it.

dnav:
Two things:
Our legals are relunctant to include the getopt code as-is in the repo (even if it is under a very permissive license).

Having put more thoughts on it, I think I'll remove the __platforms__ folder entirely and make __examples__ POSIX only (and WIN32 using your contribution with some workaround for getopt). Reasons are that having several platform ports in this repo will put more burden on the committers. Each changes would requires testing on various platforms. wakaama aims to be a reference implementation not an universal one.
If one wants to contribute a port on some platform, I would prefer it done in a specific fork (s)he would maintain than in this repo. See Julien's mbed port for instance.

Your thoughts ?

davidgraeff:
Hm, the getopt implementation is from a c compiler std lib, which is due to its nature licensed in the most permissive way. Sigh, Legals..
getopt can be clean room implemented of course. This would be less invasive to the example code than other solutions I guess.

Removing the platform directory makes sense.
Should I add a commit that moves all files from there to examples/shared?

dnav:
Allow me think on it.

jvermillard:
@dnav you have like to the CQ pointing the getops problem?

dnav:
It is from the Intel's legal, not Eclipse's one.

jvermillard:
getopt is such a common POSIX call I think it'll pass eclipse IP checks

davidgraeff:
Should I rebase this and leave the VS support out?

dnav:
I still need to remove the getopt usage. But as I do not have much time this week I wanted to merge the code reorg early.

dnav:
Hi @davidgraeff 

sorry for the dealy. You can rebase upon #146 

Regards

davidgraeff:
i'll send a rebase in the next few hours :)

Issue #136, opened 2016-04-08
bug:  dataP->id always 0 when write resource id


ieei0214:
object_write in object.c

After lwm2m_data_parse , dataP->id seems not assign any value.  So dataP->id is always 0.

Should assign "dataP->id = uriP->resourceId;"

```
coap_status_t object_write(lwm2m_context_t * contextP,
                           lwm2m_uri_t * uriP,
                           lwm2m_media_type_t format,
                           uint8_t * buffer,
                           size_t length)
{
    coap_status_t result = NO_ERROR;
    lwm2m_object_t * targetP;
    lwm2m_data_t * dataP = NULL;
    int size = 0;

    targetP = prv_findObject(contextP, uriP->objectId);
    if (NULL == targetP)
    {
        result = COAP_404_NOT_FOUND;
    }
    else if (NULL == targetP->writeFunc)
    {
        result = COAP_405_METHOD_NOT_ALLOWED;
    }
    else
    {
        size = lwm2m_data_parse(uriP, buffer, length, format, &dataP)
```
        dataP->id = uriP->resourceId;  // assign here
```
        if (size == 0)
        {
            result = COAP_406_NOT_ACCEPTABLE;
        }
    }
    if (result == NO_ERROR)
    {
        result = targetP->writeFunc(uriP->instanceId, size, dataP, targetP);
        lwm2m_data_free(size, dataP);
    }
    return result;
}
```

dnav:
Thanks for spotting this.
I made another fix: #138. The issue lies deeper in the parsing function. 

Regards

dnav:
Fixed by #138

PR #137, opened 2016-04-08
Fix for issue #136: Assign resource id when write resource.

ieei0214:
Signed-off-by: Ricky Liu <ieei.liu@gmail.com>


ieei0214:
Solved in #138

PR #142, opened 2016-04-14
Fix json opaque serialization in prv_serializeValue

davidgraeff:
utils_base64Encode writes on the wrong buffer position fixed.


dnav:
Indeed. I guess I have too many branches...

PR #144, opened 2016-04-14
Fix utils_base64Encode

davidgraeff:
The base64 encoded string is always longer than the input string:
Change the comparison operator.

Remove a not used memset in prv_encodeBlock.

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -654,8 +654,6 @@ static char b64Alphabet[64] =
 static void prv_encodeBlock(uint8_t input[3],
                             uint8_t output[4])
 {
-    memset(output, 0, 4);
-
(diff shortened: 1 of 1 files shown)


core/utils.c
```diff
@@ -726,7 +724,7 @@ size_t utils_base64Encode(uint8_t * dataP,
 
     result_len = prv_getBase64Size(dataLen);
 
-    if (result_len > dataLen) return 0;
+    if (result_len < dataLen) return 0;
```
dnav: Actually this should be a comparaison with dataLen to be sure the provided buffer is large enough. So the fix should be:
```if (result_len > bufferLen) return 0;```

davidgraeff: done

PR #147, opened 2016-04-25
improved the ASCII hex to binary key conversion.

FrankGerlach:
Signed-off-by: Frank Gerlach frank_gerlach@epam.com

### DIFF
--- a/examples/client/lwm2mclient.c
+++ b/examples/client/lwm2mclient.c
@@ -80,7 +80,6 @@
 #include <sys/stat.h>
 #include <errno.h>
 #include <signal.h>
-#include <assert.h>
 
 #define MAX_PACKET_SIZE 1024
 #define DEFAULT_SERVER_IPV6 "[::1]"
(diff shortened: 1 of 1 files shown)


examples/client/lwm2mclient.c
```diff
@@ -803,6 +805,96 @@ static void close_backup_object()
...
+             free(output);
+             return NULL;
+          }
+          input_index += 2;
+       }
+   }
+   return output;
+}
+
+void unittest_hex_2_bin()
```
dnav: This should be removed. You can put it in [wakaama]/tests.

examples/client/lwm2mclient.c
```diff
@@ -79,6 +80,7 @@
 #include <sys/stat.h>
 #include <errno.h>
 #include <signal.h>
+#include <assert.h>
```
dnav: This is not needed as it is used only in the unit test function.

examples/client/lwm2mclient.c
```diff
@@ -844,9 +938,14 @@ int main(int argc, char *argv[])
     lwm2m_client_state_t previousState = STATE_INITIAL;
 #endif
 
+
+#if 0 
+#else
```
dnav: what is the need for this ?

FrankGerlach: I turned on and off DTLS PSK with this. Of course my key material has been deleted.

FrankGerlach:
Implemented changes as requested by David Navarro.

examples/client/lwm2mclient.c
FrankGerlach: I just implemented the changes you requested.

PR #153, opened 2016-05-20
Examples: Check PSK command line argument

dogonthesun:
An information about the error is a little bit better than a segmentation fault. Segfaults make new coming users confused. To avoid this we need to provide maximum information about possible faults in examples.

I've also made some changes in the security object. Mostly I've added a few checks of pointers. I'm not sure that we can use PSK without PSK-Id. Anyway that checks allow to avoid possible segmentation faults on strdup's.

README.md is updated up to current .gitmodules file.

### DIFF
--- a/examples/client/README.md
+++ b/examples/client/README.md
@@ -17,7 +17,7 @@ git submodule update
 
 You need to install the packages libtool and autoreconf.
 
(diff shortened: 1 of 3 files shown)


dnav:
Thanks for this.

PR #154, opened 2016-05-26
dtlsconnection.c: Fix bugs in dtls connection handling

jeremy-ang:
- Fix bugs in connection context handling in send_to_peer() and read_from_peer()
- Fix a bug in ipv6 address comparison


jeremy-ang:
Found these bugs while enabling PSK DTLS support for wakaama server to handle secure connections between multiple LWM2M clients.

dnav:
Thank you

Issue #155, opened 2016-05-27
Wakaama client doesn't take new entries added to Leshan bootstrap server


kiranpradeep:
Wakaama client is successfully bootstrapping from Leshan bootstrap(BS) server and is able to  register with Leshan device management(DM) server. Now I tried below steps

1. Start Leshan BS server with no or zero entries in GUI and DM server
2. Start Wakaama client
3. Add entry corresponding to Wakaama client endpoint in Leshan BS server user interface
4. Wakaama client, during retries, doesn't receive the newly added bootstrap credentials

I also tried with Leshan client for the same scenario. But, Leshan client is able to receive the newly added credentials( waits for 600s before doing so) from Leshan bootstrap server. Is Wakaama client expected to handle this scenario ? Attaching [logs](https://github.com/eclipse/wakaama/files/286210/tkt.txt)

Leshan version: 0.1.11-M10
Wakaama: current master branch. Using lwm2mclient with DTLS disabled
Compiler: Apple LLVM version 7.3.0 (clang-703.0.29)

dnav:
Yes the wakaama client is expected to handle this.
From your logs, I can see that the client tries to reach a bootstrap server at an invalid address: ```[]:0```
I guess the objects backup and restore in the example client did not go well.
I'll investigate this.

dnav:
I reproduced your issue. With the current master, the client tries to re-connect to the right address. So I guess there is an issue with your objects backup/restore.

Regards,

kiranpradeep:
I am sorry. I couldn't understand. Is this a bug or some thing wrong with my build ? I took a checkout of wakaama master and followed it with cmake/make. My code is in sync with wakaama/master.

dnav:
Sorry I assumed you had custom components.
I gave this a second look and I found the issue (which symptoms are different depending on your platforms): when the bootstrapping fails, the lwm2mclient deletes the connection associated to the faulty server. But the core still uses it.
I'm making a quick fix for this.

kiranpradeep:
Thanks.

PR #157, opened 2016-05-30
API CHANGE: Add a new connection layer callback: lwm2m_close_connection()

dnav:
Fix issue #155.
In client mode, the core now calls  lwm2m_close_connection() at the end of the bootstrap (successful or not) and when a registration fails.


kiranpradeep:
This appears to break the build when DTLS is enabled.

dnav:
Sorry for that. Fixed.

Issue #170, opened 2016-06-10
calling prv_updateRegistration before registration was done


boria8:
if we call prv_updateRegistration before registration was done,wakaama we crash from segfault.

see the following fix code:

static int prv_updateRegistration(lwm2m_context_t * contextP,
                                  lwm2m_server_t * server,
                                  bool withObjects)
{
    lwm2m_transaction_t * transaction;
    uint8_t payload[512];
    int payload_length;

>    if(server->location == 0)
>     {
>         LOG("Server is Not registered yet\n");
>         return COAP_500_INTERNAL_SERVER_ERROR;
>     }

    transaction = transaction_new(COAP_TYPE_CON, COAP_POST, NULL, NULL, contextP->nextMID++, 4, NULL, ENDPOINT_SERVER, (void *)server);
    if (transaction == NULL) return COAP_500_INTERNAL_SERVER_ERROR;

    coap_set_header_uri_path(transaction->message, server->location);

    if (withObjects == true)
    {
        payload_length = object_getRegisterPayload(contextP, payload, sizeof(payload));
        if (payload_length == 0)
        {
            transaction_free(transaction);
            return COAP_500_INTERNAL_SERVER_ERROR;
        }
        coap_set_payload(transaction->message, payload, payload_length);
    }

    transaction->callback = prv_handleRegistrationUpdateReply;
    transaction->userData = (void *) server;

    contextP->transactionList = (lwm2m_transaction_t *)LWM2M_LIST_ADD(contextP->transactionList, transaction);

    if (transaction_send(contextP, transaction) == 0)
    {
        server->status = STATE_REG_UPDATE_PENDING;
    }

    return COAP_NO_ERROR;
}

// update the registration of a given server
int lwm2m_update_registration(lwm2m_context_t * contextP,
                              uint16_t shortServerID,
                              bool withObjects)
{
    lwm2m_server_t * targetP;
    uint8_t result;

    result = COAP_NO_ERROR;

    targetP = contextP->serverList;
    if (targetP == NULL)
    {
        if (object_getServers(contextP) == -1)
        {
            return COAP_404_NOT_FOUND;
        }
    }
    while (targetP != NULL && result == COAP_NO_ERROR)
    {
        if (shortServerID != 0)
        {
            if (targetP->shortID == shortServerID)
            {
                // found the server, trigger the update transaction

>                 if( targetP->status != STATE_REGISTERED)
>                 {
>                     LOG("Server is Not registered yet\n");
[...]

Issue #169, opened 2016-06-10
prv_refreshServerList API makes unsorted list


boria8:
prv_refreshServerList  API makes unsorted list, as a result if we have more than 2 servers, lets say instance 0,1, then instead getting 0,1 list we will get 1,0 list.
it causes wakaama to not find instance 0 server.
eventually we will have duplicated servers.

see the following  fix code:

static int prv_refreshServerList(lwm2m_context_t * contextP)
{
    lwm2m_server_t * targetP;
    lwm2m_server_t * nextP;
    lwm2m_server_t * head=0;

    // Remove all servers marked as dirty
    targetP = contextP->bootstrapServerList;
    contextP->bootstrapServerList = NULL;
    while (targetP != NULL)
    {
        DBG(DBG_VERBOSE,"bootstrapServerList not empty \n");
        nextP = targetP->next;
        if (targetP->dirty == false)
        {
            DBG(DBG_VERBOSE,"Bootstrap serverId=%d is NOT dirty, server was not deleted \n",targetP->shortID);
            targetP->status = STATE_DEREGISTERED;
            targetP->next=0;
            head = (lwm2m_server_t*)LWM2M_LIST_ADD(head, targetP);
        }
        else
        {
            DBG(DBG_VERBOSE,"prv_deleteServer \n");
            prv_deleteServer(targetP);
        }
        targetP = nextP;
    }
    contextP->bootstrapServerList=head;
    head=0;
    targetP = contextP->serverList;
    contextP->serverList = NULL;
    while (targetP != NULL)
    {
        nextP = targetP->next;
        if (targetP->dirty == false )
        {
            // TODO: Should we revert the status to STATE_DEREGISTERED ?
            DBG(DBG_VERBOSE,"serverId=%d is NOT dirty, server was not deleted  \n",targetP->shortID);
            targetP->next=0;
            head = (lwm2m_server_t*)LWM2M_LIST_ADD(head, targetP);
        }
        else
        {
            prv_deleteServer(targetP);
        }
        targetP = nextP;
    }
    contextP->serverList=head;
    return object_getServers(contextP);
}

dnav:
Thank you for spotting this.
Could you propose your fix as a Pull Request ?

Thanks,

dnav:
Actually,  I included a fix for this in #172.

Regards

dnav:
Fixed by #172

PR #172, opened 2016-06-13
Fix registration update issues.

dnav:
A failed registration update has now the same effect as a failed registration.
If there is no valid registration to a LWM2M Server, the client returns to the BOOTSTRAP_REQUIRED state.

Also include fixes for #169 and #170.

### DIFF
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -192,11 +192,11 @@ static int prv_refreshServerList(lwm2m_context_t * contextP)
     while (targetP != NULL)
     {
         nextP = targetP->next;
+        targetP->next = NULL;
         if (!targetP->dirty)
         {
             targetP->status = STATE_DEREGISTERED;
-            targetP->next = contextP->bootstrapServerList;
-            contextP->bootstrapServerList = targetP;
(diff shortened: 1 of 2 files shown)


core/liblwm2m.c
```diff
@@ -192,11 +192,11 @@ static int prv_refreshServerList(lwm2m_context_t * contextP)
     while (targetP != NULL)
     {
         nextP = targetP->next;
+        targetP->next = NULL;
```
boria8: you lose here the entire list

core/liblwm2m.c
```diff
@@ -209,11 +209,11 @@ static int prv_refreshServerList(lwm2m_context_t * contextP)
...
```
boria8: here you also lose the entire list

core/liblwm2m.c
dnav: The list is saved in `nextP` the line before, isn't it ?

core/liblwm2m.c
dnav: The list is saved in `nextP` the line before, isn't it ?

core/liblwm2m.c
boria8: you right ...

PR #176, opened 2016-06-17
duplicating psk hint and psk key.

kiranpradeep:
Duplicating string pointers. Or else this will result in double free crash. These pointers are freed later.


kiranpradeep:
Will fix the CLA warning. I had already signed Eclipse CLA.

kiranpradeep:
Closing for ip-validation error.

PR #191, opened 2016-07-15
Update management.c

JayanthSreenivasa:
For the first GET on the device /3/0/0, It can happen that the length is zero as the device holds no data, So set the message content to COAP_205_CONTENT

### DIFF
--- a/core/management.c
+++ b/core/management.c
@@ -204,7 +204,9 @@ coap_status_t dm_handleRequest(lwm2m_context_t * contextP,
                         length = lwm2m_data_serialize(uriP, size, dataP, &format, &buffer);
                         if (length == 0)
                         {
-                            result = COAP_500_INTERNAL_SERVER_ERROR;
+                            //  For the first GET on the device /3/0/0, It can happen that the length is zero as the device holds no data.
+                            //  So set the message content to COAP_205_CONTENT
+                            result = COAP_205_CONTENT;
                         }
                         else
                         {


dnav:
Hi,
this specific use case should be fixed by #187 .
Also, have a look to the [CONTRIBUTING.md ](https://github.com/eclipse/wakaama/blob/master/CONTRIBUTING.md) for instruction on the contribution requirements.
Thanks.

JayanthSreenivasa:
Hi,

Thanks for a quick reply.

But I still think there is something missing.

In the object_read function,
1. object_readData will return COAP_205_CONTENT even when there is no data
in the object. Hence the size returned will also be zero.
2. length returned by lwm2m_data_serialize will also be zero.

The check still checks is the size is not 1 (which is true) and hence
returns COAP_500_INTERNAL_SERVER_ERROR
        if (*lengthP == 0)
        {
            if (*formatP != LWM2M_CONTENT_TEXT
                *|| size != 1*
                || dataP->type != LWM2M_TYPE_STRING
                || dataP->value.asBuffer.length != 0)
            {
                result = COAP_500_INTERNAL_SERVER_ERROR;
            }
        }

Please point out if I have misunderstood something.

Thanks and Regards,
Jayanth

dnav:
If you do a Read on /3/0/0, `size` is egual to 1.
If you do a Read on /3/0 or /3, `size` will no be 1. But in this case, your device object plugin should not return the resource 0 if it is not set.
The code you are quoting is here to handle zero-length strings for single resources.

PR #211, opened 2016-10-21
add block1 support of incoming request for LWM2M client.

sbernard31:
I added the support of block1 for incoming request only for LWM2M client.

Here a list of limitation or not implemented features:
- it supports only 1 block1 transfer by server. (and not 1 transfer by server/resource URI)
- as the key/ID for block1 is the serverID, ideally we should remove pending block1 on bootstrap finished ?
- we don't check if the URI and the content format is the same from one block to another.
- we don't clean block1 buffer after EXCHANGE_LIFETIME. (we keep 1 buffer by server for retransmission)

### DIFF
--- a/core/block1.c
+++ b/core/block1.c
@@ -1,0 +1,159 @@
+/*******************************************************************************
+ *
+ * Copyright (c) 2016 Intel Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v10.html
+ * The Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * Contributors:
+ *    Simon Bernard - initial API and implementation
+ *
+ *******************************************************************************/
+/*
+ Copyright (c) 2016 Intel Corporation
+
+ Redistribution and use in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+
+     * Redistributions of source code must retain the above copyright notice,
+       this list of conditions and the following disclaimer.
+     * Redistributions in binary form must reproduce the above copyright notice,
+       this list of conditions and the following disclaimer in the documentation
+       and/or other materials provided with the distribution.
+     * Neither the name of Intel Corporation nor the names of its contributors
+       may be used to endorse or promote products derived from this software
+       without specific prior written permission.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+ IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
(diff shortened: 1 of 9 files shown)


core/liblwm2m.h
```diff
@@ -643,6 +663,7 @@ typedef struct
     uint16_t                nextMID;
     lwm2m_transaction_t *   transactionList;
     void *                  userData;
+    lwm2m_block1_data_t *   block1DataList;
```
jvermillard: should be freed in lwm2m_close() function

jvermillard: about the cleanup and to do it on bootstrap if you put your block1_data structure in the 'server' and not on the 'lwm2m_context' the bootstrap process will cleanup the server after bootstraping to new servers.
I'm not sure it's possible :)

jvermillard:
@boaks @dnav WDYT?

boaks:
I will have a look on the PR tomorrow. But I have to confess;I didn't look into wakaam for a long time ...

core/block1.c
```diff
@@ -0,0 +1,141 @@
...
+          uint8_t * oldBuffer = block1Data->block1buffer;
+          size_t oldSize = block1Data->block1bufferSize;
+          block1Data->block1bufferSize = oldSize+length;
+          block1Data->block1buffer = lwm2m_malloc(block1Data->block1bufferSize);
+          if (NULL == block1Data->block1buffer) return COAP_500_INTERNAL_SERVER_ERROR;
+          memcpy(block1Data->block1buffer, oldBuffer, oldSize);
+          lwm2m_free(oldBuffer);
+
+          // write new block in buffer
+          memcpy(block1Data->block1buffer + (blockNum * blockSize), buffer, length);
```
boaks: maybe using "oldSize" instead of "(blockNum * blockSize)" is nicer, but not really required.

core/block1.c
```diff
@@ -0,0 +1,141 @@
...
+                                  uint8_t ** outputBuffer,
+                                  size_t * outputLength)
+{
+    if (server == NULL)
+    {
+        return COAP_500_INTERNAL_SERVER_ERROR;
+    }
+
+    // find current block1 data for this server
+    lwm2m_block1_data_t * block1Data = (lwm2m_block1_data_t *) LWM2M_LIST_FIND(contextP->block1DataList, server->shortID);
```
boaks: So does this mean: 
A  block1 transfer to URI 1 is canceled by a block1 transfer to URI 2?
So you interpret "request/response" according the "complete" transfer rather then the singel requests?

sbernard31: This is the current limitation of this implementation, I mention it above :
_it supports only 1 block1 transfer by server. (and not 1 transfer by server/resource URI)_


>So you interpret "request/response" according the "complete" transfer rather then the singel requests?

Not sure to understand ? :)

core/block1.c
```diff
@@ -0,0 +1,141 @@
...
+              // we don't receive block in right order
+              // TODO should we clean block1 data for this server ?
+              return COAP_408_REQ_ENTITY_INCOMPLETE;
+          }
+
+          // re-alloc new buffer
+          uint8_t * oldBuffer = block1Data->block1buffer;
+          size_t oldSize = block1Data->block1bufferSize;
+          block1Data->block1bufferSize = oldSize+length;
+          block1Data->block1buffer = lwm2m_malloc(block1Data->block1bufferSize);
```
sbernard31: It could be an optimization. We can add it in another PR.

core/block1.c
sbernard31: You're right.

core/liblwm2m.h
sbernard31: I will add the free on `lwm2m_close`.
I will see if I can move block1_data structure in `server` instance of `lwm2m_context`.

core/block1.c
boaks: Uups, I overread the comment.

NSTART1 says something as 1 outstanding request.
With blockwise, there may be a interpretation of 
request 1 / response 1 is that outstanding request,
or requests (0-20)/responses(0-20) are that outstanding request.
The one will allow something else to happen in between, the other not.
But, as far, as I assume only a few blocks, the difference is not really worth to be discussed.

sbernard31:
I made the change.  I let 2 commit as we can choose if we want to store block1data in context or in server. (I think store it on server is better)

jvermillard:
great :)
I think if we have a #define XXXX  to configure the max block size we allow to accumulate per server we are done.

jvermillard:
Let's merge?

dnav:
> Let's merge?

Yes. No comments in 6 days since the last commits. It seems good for everyone.

PR #213, opened 2016-11-11
Implement objlink type encoding.

pasear:
Tested with the latest version of Leshan server.

Signed-off-by: wu.chunhao <wu.chunhao@nifty.co.jp>


dnav:
Looks good to me.

Issue #207, opened 2016-09-21
extra semicolon in prv_refreshServerList API


boria8:
prv_refreshServerList API has extra semicolon in the following line.
contextP->serverList = (lwm2m_server_t *)LWM2M_LIST_ADD(contextP->serverList, targetP);;

dnav:
Fixed by #218

PR #218, opened 2016-12-20
Fix for issue 217: wrong buffer offset in string copy.

### LINKED ISSUE #217: Lifetime values greater than 99999 are not well managed
If a lifetime value greater than 99999 is used, the lifetime value sent by the client to the server in REGISTER  is not correct.
For lifetime = 100000, the data sent to the server is `lt=000000`
Issue seems to be in utils.c file, in `utils_intCopy function`:
`memcpy(buffer, str + _PRV_INT32_MAX_STR_LEN - len, len);`
to be replaced by
`memcpy(buffer, str, len);`

dnav:
Fixes #217 and #207 


jvermillard:
:+1:

PR #226, opened 2017-01-16
Rehandshake on NAT timeout

sbernard31:
When a LWM2M client is behind a NAT, if it doesn't send data for a period of time, it can lost its assigned IP. This period (NAT timeout) depends of the NAT configuration.

In DTLS, the session used to decrypt packet is linked to the IP address/port of the peer. This means thats is the client lost its assigned IP, It should redo an handshake with its new IP address/port. 

This PR, based on @jvermillard code in [rehandshake-on-NAT-timeout](https://github.com/eclipse/wakaama/tree/rehandshake-on-NAT-timeout) branch,  defines a value for the NAT period, if the device is mute for `DTLS_NAT_TIMEOUT`, it will do a new handshake before to communicate again.

### DIFF
--- a/examples/shared/dtlsconnection.c
+++ b/examples/shared/dtlsconnection.c
@@ -78,7 +78,7 @@ char * security_get_public_id(lwm2m_object_t * obj, int instanceId, int * length
         dataP->type == LWM2M_TYPE_OPAQUE)
     {
         char * buff;
-        
+
         buff = (char*)lwm2m_malloc(dataP->value.asBuffer.length);
         if (buff != 0)
         {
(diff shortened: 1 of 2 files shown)


dnav:
Hi,

Looks good.
Strangely, you do not pass the Eclipse IP validation. Could you check on your side ?

Regards,

sbernard31:
I think this is a IP checker which is a bit stupid ...
You can check I sign a CLA with https://projects.eclipse.org/user/cla/validate and Julien is committer on this project. I signed the commit. I don't see any issue here.

FMPOV, the eclipse checker is just a tooling to help committer, I already merged PR without its approbation. ;)

boaks:
The "issue"may be, that @jvermillard is the committer of the first commits (even if @sbernard31 is mentioned as author), and the last ones are committed by @sbernard31. This is therefore a multi committer/author PR.

PR #227, opened 2017-01-20
update lastsend on low level socket sending instead of app-data sending

sbernard31:
This should be an enhancement of #226.
Before we update the `lastSend` each time we send Application Data. In this PR,  it's each time we write data on the socket.

### DIFF
--- a/examples/shared/dtlsconnection.c
+++ b/examples/shared/dtlsconnection.c
@@ -159,6 +159,7 @@ int send_data(dtls_connection_t *connP,
         if (nbSent == -1) return -1;
         offset += nbSent;
     }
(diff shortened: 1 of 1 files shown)


examples/shared/dtlsconnection.c
```diff
@@ -159,6 +159,8 @@ int send_data(dtls_connection_t *connP,
         if (nbSent == -1) return -1;
         offset += nbSent;
     }
+    time_t now = lwm2m_gettime();
```
jvermillard: can remove the now variable

sbernard31: Oops :p !

PR #233, opened 2017-02-16
Various fixes to small issues.

dnav:
- TLV encoding of single resource was using resource instance type instead of resource type
- Correct handling of content-type options in requests
- Check instance ID in Execute operation
- Instance ID must be set in Create operation 


dnav:
Also:
- add missing resource ID 12 of the Security Object

PR #238, opened 2017-02-17
Support old content format values (TLV/JSON)

msangoi:
Add the support of the initial/deprecated content format codes for TLV (1542) and JSON (1543) in order to keep the backward-compatibility.


boaks:
Would it be possible, to add/define a "SWITCH" to disable the "backwards-compatibility" at compile time?

dnav:
I agree with @boaks. Could you enclose your changes in a compilation switch ?
suggestion: LWM2M_OLD_CONTENT_FORMAT_SUPPORT...

msangoi:
Done.
LWM2M_OLD_CONTENT_FORMAT_SUPPORT can be used at compile time to enable the backward-compatibility for content formats. 
Disabled by default.

Issue #237, opened 2017-02-17
Object read returns an error if no object instance exist


fdur:
Each object can have one or several object instances (as indicated in LWM2M specification).
In software update specification ([object 9](http://member.openmobilealliance.org/ftp/Public_documents/DM/LightweightM2M/Permanent_documents/OMA-TS-LWM2M_SwMgmt-V1_0-20151201-C.zip)), this object can support one or several object instance.
As indicated in §6: 

> Each software component is managed via a dedicated Software Management Object instance

If no software component is managed, this object does not have any object instance.
If a READ command is received on object 9 while no object instance exist, Wakaama returns `COAP_404_NOT_FOUND`
From my point of view, in this case, Wakaama needs to return `COAP_205_CONTENT`with empty payload

dnav:
I tend to share your point of view.
I'll investigate this.

dnav:
See #240

dnav:
Fixed by #240

dnav:
Reopen following @jvermillard comment:
> This fix doesn't work because the content format is NULL so the server can't parse the payload as a TLV.
Also if the server read as JSON wwe should have "{}" and not an empty payload. IMO We should add the @fdur fix

dnav:
Hi,
several aspects of the issue:
1- the content type should not be NULL if the LwM2M Server sent an accept option. I'll investigate this.
2- IMHO an empty payload is a valid TLV payload
3- You're right, the JSON payload should be "{}"

jvermillard:
For 2 & 3 I think we should manage an empty instance list as any other content and encode it and don't try to make a special case

dnav:
You're right, there are too much special cases now.
It will be easier to manage by changing the returned value of lwm2m_data_serialize() so that 0 is now longer an error.

dnav:
Please have a look at #251 
For empty Objects, the JSON payload is `{"bn":"/1234/","e":[]}`.
The TLV payload is an empty payload.
The right content-type is returned correctly.

PR #240, opened 2017-02-20
Fix for issue #237: report empty payload for Read on /ObjectID for Ob…

dnav:
…jects with no Instances.

Also handle Discover operation.

### DIFF
--- a/core/data.c
+++ b/core/data.c
@@ -522,7 +522,7 @@ size_t lwm2m_data_serialize(lwm2m_uri_t * uriP,
 
 #ifdef LWM2M_CLIENT_MODE
     case LWM2M_CONTENT_LINK:
-        return discover_serialize(NULL, uriP, size, dataP, bufferP);
+        return discover_serialize(NULL, uriP, NULL, size, dataP, bufferP);
 #endif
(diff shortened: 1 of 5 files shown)


dnav:
Piggybacking: align attributes reporting with the latest TS.

fdur:
Hi,
in your updates, I think that in objects.c file, in object_read API, the following update is needed:

From
```
if (result == COAP_205_CONTENT)
    {
        if (size == 0)
        {
            *lengthP = 0;
        }
```
To
```
if (result == COAP_205_CONTENT)
    {
        if (size == 0)
        {
            *lengthP = lwm2m_data_serialize(uriP, size, dataP, formatP, bufferP);
        }
```
in order to get the right format (else format seems to remains NULL and not set to LWM2M_CONTENT_TLV or JSON)

dnav:
Hi,

In another change already merged into master but ot in this branch, before calling object_read(), format is set to the value in the accept option or `text format` if this one is not present.
I think this solves the issue you are pointing.

Regards,

jvermillard:
This fix doesn't work because the content format is NULL so the server can't parse the payload as a TLV

PR #251, opened 2017-03-14
Fix #237 v2

### LINKED ISSUE #237: Object read returns an error if no object instance exist

dnav:
API Change: lwm2m_data_serialize() now returns an integer. 0 is a valid value, -1 is an error.

### DIFF
--- a/core/data.c
+++ b/core/data.c
@@ -21,25 +21,33 @@
 #include <float.h>
 
 // dataP array length is assumed to be 1.
-static size_t prv_textSerialize(lwm2m_data_t * dataP,
-                                uint8_t ** bufferP)
+static int prv_textSerialize(lwm2m_data_t * dataP,
+                             uint8_t ** bufferP)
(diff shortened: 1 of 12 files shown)


jvermillard:
LGTM :+1:

Issue #266, opened 2017-04-15
lwm2m_data_serialize() doesn't consider older json/tlv formats 


kiranpradeep:
It appears current master([1](https://github.com/eclipse/wakaama/blob/86b41ddd0f9a9a03b4ab52881b563044bc722861/core/data.c#L471)) doesn't take  `LWM2M_*_OLD` into account while serializing data.

I tried modifying  `lwm2m_data_serialize` and read was succeeding. But then observations failed.

Is this(#238) now a non goal of this project ?

dnav:
Hi,

The fix for #238 was incomplete. Please have a look to #267.
The accept option was also ignored in Observation requests.

PR #267, opened 2017-04-18
Fix for issue #266: Handle JSON and TLV old content format values

dnav:
when encoding data to match the accept option of the request.
Also consider the accept option in observe requests.


kiranpradeep:
@dnav with leshan.eclipse.org, I tried observing battery( lwm2mclient -c), on this fix_266 branch. It didn't work. Observation are good on master though.

dnav:
Indeed. I forgot a change.
It should be OK now.

PR #293, opened 2017-05-22
Remove redundant assignments

scop:
Signed-off-by: Ville Skyttä <ville.skytta@iki.fi>


dnav:
Hi,

There may be side-effect. Following the wise old saying "If it's not broken, don't fix it.", I am relunctant to merge this.

Regards,

Issue #257, opened 2017-03-22
Bootstrap leaves connection open on client.


sbertin-telular:
The bootstrap_step function with status STATE_BS_FINISHED would close the connection, but it is never called with that status. The lwm2m_step function in the state STATE_BOOTSTRAPPING calls bootstrap_getStatus and only calls bootstrap_step if the status is not STATE_BS_FINISHED and not STATE_BS_FAILED.
	
I would suggest having the lwm2m_step function in the state STATE_BOOTSTRAPPING call the bootstrap_step function in all cases. I have tested moving it from the default case to before the switch statement and it disconnected properly.

dnav:
Indeed.
Your suggestion fixes the issue but I'll investigate to determine if a design change would be better.
Thanks

sbertin-telular:
That change looks like it will handle the successful case, but I think the connection will still be left open in the failed case.

dnav:
You're right. It should be OK now.

sbertin-telular:
bootstrap_getStatus should return STATE_BS_PENDING in STATE_BS_FAILING. It currently returns STATE_BS_FAILED, but that prevents bootstrap_step for being called to close the connection and move from STATE_BS_FAILING to STATE_BS_FAILED.

dnav:
Fixed by #294

PR #294, opened 2017-05-24
Add missing failing case to fix issue #257

sbertin-telular:


dnav:
With this change, BS_FAILING is reported as BS_PENDING. Currently, it is reported as BS_FAILED. I can not see the issue. Could you elaborate ?

Thanks,

sbertin-telular:
Returning BS_FAILED prevents bootstrap_step from being called to close the connection and move from STATE_BS_FAILING to STATE_BS_FAILED.

dnav:
Indeed. Issue #257. Sorry.

Issue #289, opened 2017-05-19
IANA code for TLV


fdur:
2 IANA codes are support for TLV:
```
typedef enum
{
    LWM2M_CONTENT_TEXT      = 0,        // Also used as undefined
    LWM2M_CONTENT_LINK      = 40,
    LWM2M_CONTENT_OPAQUE    = 42,
    LWM2M_CONTENT_TLV_OLD   = 1542,     // Keep old value for backward-compatibility
    LWM2M_CONTENT_TLV       = 11542,
    LWM2M_CONTENT_JSON_OLD  = 1543,     // Keep old value for backward-compatibility
    LWM2M_CONTENT_JSON      = 11543
} lwm2m_media_type_t;
```

The OLD value (1542) is supported if LWM2M_OLD_CONTENT_FORMAT_SUPPORT compilation flag is used.
If the Wakaama Linux client connect to a server and received a READ command on /3 (example), even if the LWM2M_OLD_CONTENT_FORMAT_SUPPORT compilation flag is used, the coding value for content-type is the new one

Example
```May 19 10:37:55 DEBG send unencrypted: (160 bytes):
00000000 67 45 89 11 EE 2C 3C CD  59 36 7D C2 2D 16 FF 08 
00000010 00 8E C8 00 14 4F 70 65  6E 20 4D 6F 62 69 6C 65 
00000020 20 41 6C 6C 69 61 6E 63  65 C8 01 16 4C 69 67 68 
00000030 74 77 65 69 67 68 74 20  4D 32 4D 20 43 6C 69 65 
00000040 6E 74 C8 02 09 33 34 35  30 30 30 31 32 33 C3 03 
00000050 31 2E 30 86 06 41 00 01  41 01 05 88 07 08 42 00 
00000060 0E D8 42 01 13 88 87 08  41 00 7D 42 01 03 84 C1 
00000070 09 64 C1 0A 0F 83 0B 41  00 00 C8 0D 08 00 00 00 
00000080 00 AA A0 F1 72 C6 0E 2B  30 31 3A 30 30 C8 0F 0D 
00000090 45 75 72 6F 70 65 2F 42  65 72 6C 69 6E C1 10 55
```
1st line,: 2D 16
Better to send in this case 06 06 (1542)
Old value is not supported in `lwm2m_data_serialize`

dnav:
For me, the LWM2M_OLD_CONTENT_FORMAT_SUPPORT means that wakaama handles **both** the new and the old values.
[...]

PR #297, opened 2017-05-30
Fix for #289: IANA code for TLV

fdur:
Signed-off-by: Frederic DUR <fdur@sierrawireless.com>


dnav:
Hi,

please have a look to my comment in #289.

Regards

fdur:
OK
From my point of view, I thought that the LWM2M_OLD_CONTENT_FORMAT_SUPPORT flag meant that only old code for TLV/JSON are supported.

Your remark makes sense: close the ticket and the pull request

Issue #301, opened 2017-05-31
Infinite loop in server after observing the same thing twice


sbertin-telular:
This is due to a problem in lwm2m_observe. Adding an observation to the server first looks for an already existing observation that matches. This will then add the existing observation to the observation list a second time. This results in an infinite loop when traversing the list later, as the observation will point to itself for the next observation.

I see three potential solutions:
1. Avoid adding the observation to the list in lwm2m_observe if it was found in the list.
2. Modify lwm2m_list_add to avoid adding something to the list that is already there. This would have the benefit of preventing the problem for all lists.
3. Return an error from lwm2m_observe for a duplicate observation instead of processing.

I would suggest both 1 and 2, but wanted to solicit opinions before I create a pull request.

dnav:
Solution 1) is the simplest.
Actually, it is so simple that I took the liberty to create a pull request while I was reviewing the issue. See #302.

Regards,

dnav:
Fixed by #302

PR #302, opened 2017-06-01
Fix for issue #301: Do not re-enqueue the observation to the list…

dnav:
…to avoid a loop.

### DIFF
--- a/core/observe.c
+++ b/core/observe.c
@@ -912,6 +912,8 @@ int lwm2m_observe(lwm2m_context_t * contextP,
         observationP->id = lwm2m_list_newId((lwm2m_list_t *)clientP->observationList);
         memcpy(&observationP->uri, uriP, sizeof(lwm2m_uri_t));
         observationP->clientP = clientP;
+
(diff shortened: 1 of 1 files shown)


core/observe.c
```diff
@@ -925,12 +927,11 @@ int lwm2m_observe(lwm2m_context_t * contextP,
     transactionP = transaction_new(clientP->sessionH, COAP_GET, clientP->altPath, uriP, contextP->nextMID++, 4, token);
     if (transactionP == NULL)
     {
+        observationP->clientP->observationList = (lwm2m_observation_t *)LWM2M_LIST_RM(observationP->clientP->observationList, observationP->id, NULL);
```
sbertin-telular: I'm not sure the logic is quite right here. An existing observation can be removed because the transaction to send the observe again can't be created. The client could still be sending notifies, but the server can no longer process them.

dnav: I agree the existing registration may be removed. But in the first place you decided to modify it. So in case of failure, everything is removed. I
If the client sends a notification for a non-existing observation, the server replies with a RST message cancelling the observation.

PR #318, opened 2017-07-04
Fix compiler warnings if compiled with -Wall -pedantic

davidgraeff:


dnav:
Hi,
there was a conflict in registration.c so maybe there is still a warning in this file.

Regards,

Issue #320, opened 2017-07-06
data serialize bug！


huwentao1222:
if  a server want to GET a resource which is  integer type or float .etc and Accept type is OPAQUE,  lwm2m_data_serialize function will malloc memory of undifined size and copy undefined value while dataP->type is LWM2M_TYPE_INTEGER.see below:
```C
 switch (*formatP)
    {
    case LWM2M_CONTENT_TEXT:
        return prv_textSerialize(dataP, bufferP);

    case LWM2M_CONTENT_OPAQUE:
        *bufferP = (uint8_t *)lwm2m_malloc(dataP->value.asBuffer.length);
        if (*bufferP == NULL) return -1;
        memcpy(*bufferP, dataP->value.asBuffer.buffer, dataP->value.asBuffer.length);
        return (int)dataP->value.asBuffer.length;
```

dnav:
Thank you for reporting this.

Please see #321 as a proposed fix.

Regards,

huwentao1222:
Is there a need to write a function `prv_opaqueSerialize` just like `prv_textSerialize`? Maybe it is friendly to users to write object related functions. However, It's also OK of this fix.

huwentao1222:
```c
if (*formatP == LWM2M_TYPE_OPAQUE
     && dataP->type != LWM2M_TYPE_OPAQUE)
    {
        LOG("Opaque format is reserved to opaque resources.");
        return -1;
    }
```
Is `LWM2M_TYPE_OPAQUE` should be `LWM2M_CONTENT_OPAQUE`?

huwentao1222:
Is there any need to write a function `prv_opaqueSerialize` just like `prv_textSerialize`?

dnav:
Ï fixed the bad enum.

Currently there are no need for a prv_opaqueSerialize function as only opaque data can be encoded as opaque. This may change with future versions of LwM2M.

Regards,

dnav:
Fixed by #321

PR #321, opened 2017-07-06
Fix for #320: only opaque data can be encoded in opaque.

dnav:
Also introduce text encoding of opaque data.

Signed-off-by: David Navarro <david.navarro@ioterop.com>


huwentao1222:
Is there a need to write a function `prv_opaqueSerialize` just like `prv_textSerialize`? Maybe it is friendly to users to write object related functions. However, It's also OK of this fix.

PR #338, opened 2017-10-31
Avoid accessing freed memory when deleting object

Saku2:
Calling deleteFunc for object instance will invalidate the instance
pointer. Store instanceId before calling deleteFunc for object instance
to avoid use after free.


dnav:
Looks good to me. Thank you.

PR #343, opened 2017-11-24
Correct step CoRE Link param as "st"

rbtchc:
Per OMA-TS-LightweightM2M-V1_0_1-20170704-A, section 5.1.2, table 4.
The step attribute Core link param should be "st" instead of "stp"

Signed-off-by: Robert Chou <robert.ch.chou@acer.com>


dnav:
Looks good. Thank you.

PR #348, opened 2017-12-20
Change object_checkReadable function.

asarabi:
object_checkReadable and object_checkNumeric code are duplicated.
When it called, it have to call readFunc to check readable. 
If object_checkReadable have attr argument, It could check resource is numeric or not.

Signed-off-by: Changkeun IM <asarabi8282@gmail.com>

### DIFF
--- a/core/objects.c
+++ b/core/objects.c
@@ -18,7 +18,7 @@
  *    Bosch Software Innovations GmbH - Please refer to git log
  *    Pascal Rieux - Please refer to git log
  *    Scott Bertin - Please refer to git log
- *    
+ *
  *******************************************************************************/
 
 /*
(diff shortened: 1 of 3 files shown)


core/objects.c
```diff
@@ -85,48 +86,22 @@ uint8_t object_checkReadable(lwm2m_context_t * contextP,
...
-            break;
-        default:
-            result = COAP_405_METHOD_NOT_ALLOWED;
+            switch (dataP->type)
+            {
+                case LWM2M_TYPE_INTEGER:
+                case LWM2M_TYPE_FLOAT:
+                    break;
+                default:
+                    result = COAP_405_METHOD_NOT_ALLOWED;
```
sbertin-telular: Missing a break statement for the default case,

asarabi: Thanks for your review.

PR #358, opened 2018-02-20
Fix transaction userData for deregistration.

sbertin-telular:
prv_handleDeregistrationReply() was expecting the transaction userData
to be the server, but it was being set to the context in
registration_deregister().


dnav:
Thank you for this.

PR #359, opened 2018-02-21
Fix issues found cancelling observes

sbertin-telular:
There were issues where the wrong observation would be cancelled or the observation wouldn't be canceled at all.


dnav:
Hi,

would you have the steps to reproduce the issue ?

Thanks

PR #372, opened 2018-06-12
Add support for CoAP Block1 write while sending large registration an…

SushainRazdan:
…d update strings

Signed-off-by: Sushain Razdan <razz.sushain@gmail.com>

### DIFF
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -69,6 +69,8 @@ lwm2m_context_t * lwm2m_init(void * userData)
         contextP->userData = userData;
         srand((int)lwm2m_gettime());
         contextP->nextMID = rand();
+        contextP->remaining_dataLen = 0;
+        contextP->block_no = 0;
     }
 
     return contextP;
@@ -441,11 +443,14 @@ next_step:
             break;
 
         case STATE_REG_FAILED:
+            contextP->remaining_dataLen = 0;
+            contextP->block_no = 0;
+        case STATE_REG_PARTIAL_DONE:
             // TODO avoid infinite loop by checking the bootstrap info is different
             contextP->state = STATE_BOOTSTRAP_REQUIRED;
             goto next_step;
             break;
-
+        case STATE_REG_PARTIAL_PROGRESS:
         case STATE_REG_PENDING:
         default:
             // keep on waiting
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -69,8 +69,6 @@ lwm2m_context_t * lwm2m_init(void * userData)
         contextP->userData = userData;
         srand((int)lwm2m_gettime());
         contextP->nextMID = rand();
-        contextP->remaining_dataLen = 0;
-        contextP->block_no = 0;
     }
 
     return contextP;
@@ -441,16 +439,10 @@ next_step:
         case STATE_REGISTERED:
             contextP->state = STATE_READY;
             break;
-
         case STATE_REG_FAILED:
-            contextP->remaining_dataLen = 0;
-            contextP->block_no = 0;
-        case STATE_REG_PARTIAL_DONE:
-            // TODO avoid infinite loop by checking the bootstrap info is different
             contextP->state = STATE_BOOTSTRAP_REQUIRED;
             goto next_step;
             break;
-        case STATE_REG_PARTIAL_PROGRESS:
         case STATE_REG_PENDING:
         default:
             // keep on waiting
--- a/core/liblwm2m.h
+++ b/core/liblwm2m.h
@@ -475,6 +475,8 @@ typedef struct _lwm2m_server_
     char *                  location;
     bool                    dirty;
     lwm2m_block1_data_t *   block1Data;   // buffer to handle block1 data, should be replace by a list to support several block1 transfer by server.
+    int block_offset;
+    uint16_t block_no;
 } lwm2m_server_t;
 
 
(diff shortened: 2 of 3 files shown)


core/liblwm2m.h
```diff
@@ -667,6 +671,8 @@ typedef struct
     uint16_t                nextMID;
     lwm2m_transaction_t *   transactionList;
     void *                  userData;
+    int remaining_dataLen;
```
sbertin-telular: To allow registration with multiple servers I think this needs to be in lwm2m_server_t instead.

core/liblwm2m.c
```diff
@@ -441,11 +443,14 @@ int lwm2m_step(lwm2m_context_t * contextP,
             break;
 
         case STATE_REG_FAILED:
+            contextP->remaining_dataLen = 0;
+            contextP->block_no = 0;
+        case STATE_REG_PARTIAL_DONE:
```
sbertin-telular: Why would this cause failure? Seems to me it should send the next block. Maybe I'm just missing where the next block gets sent.

core/registration.c
```diff
@@ -526,7 +572,11 @@ lwm2m_status_t registration_getStatus(lwm2m_context_t * contextP)
             case STATE_REG_PENDING:
                 reg_status = STATE_REG_PENDING;
                 break;
-
+            case STATE_REG_PARTIAL_PROGRESS:
```
sbertin-telular: Seems to me this should just return STATE_REG_PENDING.

sbertin-telular:
Overall I'm glad to see this. I may be needing it soon.

core/registration.c
SushainRazdan: I have added two registration states viz.
1. STATE_REG_PARTIAL_PROGRESS: This indicates that a block of data has been sent, but 2.31 Continue response is yet to be received from the server

2. STATE_REG_PARTIAL_DONE: This would indicate that 2.31 response has been received for that block.

During block transfer registration will oscillate between these states for multiple blocks till the last block is sent where the state changes to STATE_REG_PENDING

SushainRazdan:
:)

core/liblwm2m.c
SushainRazdan: There are 2 cases.
1. When there is a failure in block transfer of registration string and we need to reset the block_no and remaining_dataLen.
2. When there is blocks have been partially transferred.

In either case, registration_start needs to be called and based on block_no and remaining_dataLen, the respective block data gets transferred.

sbertin-telular: It looks like case 2 would start a bootstrap rather than sending the next block. I think it should go to STATE_REGISTER_REQUIRED to call registration_start rather than STATE_BOOTSTRAP_REQUIRED.

core/registration.c
```diff
@@ -211,6 +212,10 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
 
             LOG("Registration successful");
         }
+        else if (packet != NULL && packet->code == COAP_231_CONTINUE)
```
sbertin-telular: RFC 7959 allows the final response to come before all blocks have been sent. It is probably better to look at the Block1 option and how much has been sent before processing based on packet->code. It would also be good to check the size in the Block1 option and reduce the block size if requested.

core/registration.c
```diff
@@ -278,11 +286,26 @@ static uint8_t prv_register(lwm2m_context_t * contextP,
         lwm2m_free(query);
         return COAP_503_SERVICE_UNAVAILABLE;
     }
-
+    if (contextP->remaining_dataLen > MAX_BLOCK1_SIZE)
```
sbertin-telular: I'd like to see the size determination for whether or not to use Block1 separate from the Block1 size. For example, it might be possible to send a 1400 byte payload in a single message without using Block1. Block1 support is also optional, so it could prevent registering with a server that would succeed with a slightly larger message.

core/registration.c
sbertin-telular: I understand that, I just think it shouldn't be exposed to the calling functions. I think it can be handled entirely within the registration processing. Either the reply can trigger sending the next block or it could be done in registration_step.

If it stays this way, the comment just above the start of registration_getStatus needs to be updated to indicate the new return values.

core/liblwm2m.c
SushainRazdan: Point noted, will change this

core/registration.c
SushainRazdan: Okay, I will look at this.

core/registration.c
SushainRazdan: Thanks, and agree, it can be done without exposing it to liblwm2m. Will check this out too.

core/liblwm2m.h
SushainRazdan: Okay.

core/registration.c
SushainRazdan: I am sorry, but this is not clear to me. How do we decide when to use block transfer ? Do we initiate a transfer of 1400 bytes in one go and if it fails, fallback to block transfer ?

sbertin-telular: I was just thinking a separate #define to determine whether or not to use Block1. If the total payload size is greater, then use Block1. Otherwise send it in one go. Make sure both that and MAX_BLOCK1_SIZE can be overridden by a define on the command line. The Internet generally has an MTU about 1500 bytes, but something like 6LoWPAN would be much smaller.

SushainRazdan:
There have been some issues with my ECA, Hence closing this pull request

PR #373, opened 2018-06-15
Fix 258: Add support for CoAP Block1 write while sending large registration an…

SushainRazdan:
…d update strings.

Signed-off-by: Sushain Razdan <razz.sushain@gmail.com>

### DIFF
--- a/core/registration.c
+++ b/core/registration.c
@@ -222,47 +222,48 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
                 uint16_t block1_size;
                 uint32_t block1_offset;
                 coap_get_header_block1(packet, &block1_num, &block1_more, &block1_size, &block1_offset);
-                targetP->rem_payload_length -= block1_size;
-                if (block1_more == 0) targetP->final_block_response = true;
+                targetP->rem_payload_length -= targetP->block_size;
 
-                if (packet != NULL && (packet->code == COAP_201_CREATED || packet->code == COAP_231_CONTINUE))
+                if (packet->code == COAP_201_CREATED && block1_more == 0)
                 {
-                    if (packet->code == COAP_201_CREATED)
+                    if (NULL != targetP->location)
                     {
-                        if (NULL != targetP->location)
-                        {
-                            lwm2m_free(targetP->location);
-                        }
-                        targetP->location = coap_get_multi_option_as_string(packet->location_path);
+                        lwm2m_free(targetP->location);
                     }
-                    /* All blocks have been sent */
-                    if (targetP->rem_payload_length <= 0)
+                    targetP->location = coap_get_multi_option_as_string(packet->location_path);
+                    targetP->final_block_response = true;
+                }
+                /* All blocks have been sent */
+                if (targetP->rem_payload_length <= 0)
+                {
+                    /* registration is successful only if we had got a final response in
+                     * any of the block transfer requests */
+                    if (targetP->final_block_response)
                     {
-                        /* registration is successful only if we had got a final response in
-                         * any of the block transfer requests */
-                        if (targetP->final_block_response)
-                        {
-                            targetP->status = STATE_REGISTERED;
-                            LOG("Registration successful");
-                        }
-                        else
-                        {
-                            targetP->status = STATE_REG_FAILED;
-                            LOG("Registration failed");
-                        }
-                        targetP->block_no = 0;
-                        targetP->block_size = 0;
-                        targetP->rem_payload_length = 0;
-                        targetP->final_block_response = false;
+                        targetP->status = STATE_REGISTERED;
+                        LOG("Registration successful");
                     }
+                    else
+                    {
+                        targetP->status = STATE_REG_FAILED;
+                        LOG("Registration failed");
+                    }
+                    targetP->block_no = 0;
+                    targetP->block_size = 0;
+                    targetP->rem_payload_length = 0;
+                    targetP->final_block_response = false;
                 }
                 if (targetP->rem_payload_length > 0)
                 {
                     if (block1_size <= targetP->block_size)
                     {
                         targetP->block_no += targetP->block_size/block1_size;
+                        targetP->block_size = block1_size;
                     }
-                    targetP->block_size = block1_size;
+                    else
+                    {
+                        targetP->block_no++;
+                    }
                     prv_register(contextP, targetP);
                 }
             }
@@ -413,7 +414,7 @@ static void prv_handleRegistrationUpdateReply(lwm2m_transaction_t * transacP,
         {
             targetP->registration = tv_sec;
         }
-        if (packet != NULL && IS_OPTION(packet, COAP_OPTION_BLOCK1) && (packet->code & 0xE0) == 2 << 5)
+        if (packet != NULL && IS_OPTION(packet, COAP_OPTION_BLOCK1))
         {
             if ((packet->code & 0xE0) == 2 << 5)
             {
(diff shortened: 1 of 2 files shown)


core/registration.c
```diff
@@ -193,8 +194,21 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
...
+        if (IS_OPTION(packet, COAP_OPTION_BLOCK1))
+        {
+            uint32_t block1_num;
+            uint8_t  block1_more;
+            uint16_t block1_size;
+            uint32_t block1_offset;
+            coap_get_header_block1(packet, &block1_num, &block1_more, &block1_size, &block1_offset);
+            if (block1_more)
+            {
+                targetP->block_offset = block1_offset + block1_size;
```
sbertin-telular: block1_more just indicates whether or not the final response to the request is included. Any remaining blocks still need to be sent (as long as the response code is 2,xx). The logic isn't right with a block size change either. Suppose block 0 is sent with a size of 1024 and the response comes back with a block size of 512. Changing to the smaller block size means block 1 was already sent with block 0 and the next block sent should be block 2 with a size of 512.

core/registration.c
```diff
@@ -211,6 +225,10 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
 
             LOG("Registration successful");
         }
+        else if (packet != NULL && packet->code == COAP_231_CONTINUE)
```
sbertin-telular: There is the potential that COAP_201_CREATED would come back before the final block is sent. In that case, the location should be stored, but the targetP->status not change to STATE_REGISTERED until the response to the final block.

sbertin-telular: In the case where the COAP_201_CREATED comes back before the final block, the final block would have a COAP_231_CONTINUE response code. In this case there are no more blocks to send and targetP->status should be set to STATE_REGISTERED.

core/registration.c
sbertin-telular: block1_size would also need to be saves and used when setting the block1 option in place of MAX_BLOCK1_SIZE.

core/registration.c
```diff
@@ -61,6 +61,7 @@
 #include <stdio.h>
 
 #define MAX_LOCATION_LENGTH 10      // strlen("/rd/65534") + 1
+#define MAX_BLOCK1_SIZE 1024
```
sbertin-telular: Would like to see:
#ifndef MAX_BLOCK1_SIZE
#define MAX_BLOCK1_SIZE 1024
#endif
so that it can be overridden for a specific project to select a smaller size.

core/registration.c
```diff
@@ -1330,16 +1396,21 @@ void registration_step(lwm2m_context_t * contextP,
             }
         }
         break;
-
+        case STATE_REG_PARTIAL_DONE:
+            prv_register(contextP, targetP);
```
sbertin-telular: In the example client, this function may only get called every 60s. That seems a long time to wait between messages, and would place an upper limit of 4 blocks withing the EXCHANGE_LIFETIME. Can sending the next block be triggered by receipt of the response instead?

sbertin-telular: One alternative might be to set *timeoutP to 1 (or 2, the CoAP ACK_TIMEOUT) while the blocks are being sent. Battery powered devices may have tradeoffs between battery life, the frequency of checking, and the total time they need to have a radio powered on. Triggering the next send on receipt of the response would likely be best for battery life.

SushainRazdan: Sorry for delayed response, As per my understanding in order to trigger next block transfer we need to call prv_register() for which we need a pointer to lwm2m_context, which is not there in handle_RegistrationReply(). Is there any implementation which comes to your mind ?

sbertin-telular: I hadn't considered that, I'm actually using a modified version that does pass in the context. Modifying the transaction callback to pass the context is one option, but isn't backward compatible.

If transaction->userData is set to the context instead of the server, the server could be determined by looking for a server with a sessionH matching transaction->peerH (use lwm2m_session_is_equal).

SushainRazdan: There is a utils_findServer function, in utils.c, I think it does the same as what you mentioned.

core/registration.c
sbertin-telular: Registration update could return 2.04 changed. To easily check for any success code, check if (packet->code & 0xE0) == (2 << 5)

core/registration.c
```diff
@@ -191,30 +199,102 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
...
+                        targetP->final_block_response = false;
+                    }
+                }
+                if (targetP->rem_payload_length > 0)
+                {
+                    if (block1_size <= targetP->block_size)
+                    {
+                        targetP->block_no += targetP->block_size/block1_size;
+                    }
+                    targetP->block_size = block1_size;
```
sbertin-telular: This should be within the if. Block size should never increase.

core/registration.c
```diff
@@ -191,30 +199,102 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
...
+                        targetP->block_size = 0;
+                        targetP->rem_payload_length = 0;
+                        targetP->final_block_response = false;
+                    }
+                }
+                if (targetP->rem_payload_length > 0)
+                {
+                    if (block1_size <= targetP->block_size)
+                    {
+                        targetP->block_no += targetP->block_size/block1_size;
```
sbertin-telular: Missing targetP->block_no update when block1_size > targetP->block_size. Should just go to the next block number.

core/registration.c
```diff
@@ -191,30 +199,102 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
...
+            /* we should send next blocks only if we get success */
+            if ((packet->code & 0xE0) == 2 << 5)
             {
-                lwm2m_free(targetP->location);
+                uint32_t block1_num;
+                uint8_t  block1_more;
+                uint16_t block1_size;
+                uint32_t block1_offset;
+                coap_get_header_block1(packet, &block1_num, &block1_more, &block1_size, &block1_offset);
+                targetP->rem_payload_length -= block1_size;
```
sbertin-telular: I think this should subtract targetP->block_size instead of block1_size. That represents how much was sent.

core/registration.c
```diff
@@ -306,24 +401,85 @@ static void prv_handleRegistrationUpdateReply(lwm2m_transaction_t * transacP,
...
         if (tv_sec >= 0)
         {
             targetP->registration = tv_sec;
         }
-        if (packet != NULL && packet->code == COAP_204_CHANGED)
+        if (packet != NULL && IS_OPTION(packet, COAP_OPTION_BLOCK1) && (packet->code & 0xE0) == 2 << 5)
         {
-            targetP->status = STATE_REGISTERED;
-            LOG("Registration update successful");
+            if ((packet->code & 0xE0) == 2 << 5)
```
sbertin-telular: This is already checked in the if just above.

core/registration.c
```diff
@@ -191,30 +199,102 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
...
             {
-                lwm2m_free(targetP->location);
+                uint32_t block1_num;
+                uint8_t  block1_more;
+                uint16_t block1_size;
+                uint32_t block1_offset;
+                coap_get_header_block1(packet, &block1_num, &block1_more, &block1_size, &block1_offset);
+                targetP->rem_payload_length -= block1_size;
+                if (block1_more == 0) targetP->final_block_response = true;
```
sbertin-telular: Should probably also check packet->code == COAP_201_CREATED. Nothing else indicates a final successful response.

core/registration.c
```diff
@@ -191,30 +199,102 @@ static void prv_handleRegistrationReply(lwm2m_transaction_t * transacP,
...
+
+                if (packet != NULL && (packet->code == COAP_201_CREATED || packet->code == COAP_231_CONTINUE))
```
sbertin-telular: I think just check if packet != null. As it is, if the last packed has an unexpected successful code, the registration will never go to STATE_REGISTERED or STATE_REG_FAILED.

sbertin-telular:
#461 looks to be a more general solution. Should this PR be closed without merging?

sbernard31:
@tuve could you confirm we can close this one ?

tuve:
Yes, this is included in #461

PR #375, opened 2018-06-19
Update json.c

### LINKED ISSUE #371: Json serialization of empty opaque values
An OPAQUE resource is allowed to be empty, isn't it? If it is, the current code handles this case wrong.

In `json.c` in `prv_serializeValue` line 1122:
```c++
    case LWM2M_TYPE_OPAQUE:
        ...
        res = utils_base64Encode(tlvP->value.asBuffer.buffer, tlvP->value.asBuffer.length, buffer+head, bufferLen - head);
        if (res == 0) return -1;
```
The if condition should be removed. An empty opaque buffer should just result in an empty base64 string.

davidgraeff:
Fixes #371

### DIFF
--- a/core/json.c
+++ b/core/json.c
@@ -1120,7 +1120,6 @@ static int prv_serializeValue(lwm2m_data_t * tlvP,
         head = JSON_ITEM_STRING_BEGIN_SIZE;
 
         res = utils_base64Encode(tlvP->value.asBuffer.buffer, tlvP->value.asBuffer.length, buffer+head, bufferLen - head);
-        if (res == 0) return -1;
         head += res;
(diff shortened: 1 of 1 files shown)


dnav:
With this change, we do not detect encoding errors. Would it be possible to test if the `tlvP->value.asBuffer.length` is 0 or not ?

For being able to merge your PR you need to update your Eclipse CLA to match your commit email address. See: https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Foundation_Contributor_License_Agreement

Regards,

dnav:
Addressed in #390

PR #376, opened 2018-06-19
Update json.c

### LINKED ISSUE #370: json serialize: Doesn't propagate error
In `json.c` in `json_serialize(...)` line 1416:
```
    for (index = 0 ; index < num && head < PRV_JSON_BUFFER_SIZE ; index++) {
        int  res = prv_serializeData(targetP + index, NULL, 0, bufferJSON + head, PRV_JSON_BUFFER_SIZE - head);
        if (res < 0) return 0;
        head += res;
    }
```
If `prv_serializeData` fails, 0 is returned. This is not identified as an error in the calling function (lwm2m_data_serialize <- object_read), but a COAP_500_INTERNAL_SERVER_ERROR should be returned.

Wouldn't it be more correct if all erroneous code paths in `json_serialize(...)` return `-1` instead of `0`?

davidgraeff:
Fixes #370

### DIFF
--- a/core/json.c
+++ b/core/json.c
@@ -1414,7 +1414,7 @@ int json_serialize(lwm2m_uri_t * uriP,
         int res;
 
         res = prv_serializeData(targetP + index, NULL, 0, bufferJSON + head, PRV_JSON_BUFFER_SIZE - head);
-        if (res < 0) return 0;
+        if (res < 0) return -1;
         head += res;
     }
 


dnav:
Thank you for this.

For being able to merge your PR you need to update your Eclipse CLA to match your commit email address. See: https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Foundation_Contributor_License_Agreement

Regards,

davidgraeff:
One line changes do not need the automatic check to pass. A correct signed-off-by is enough. That's all least how other eclipse projects handle it. 

The problem: the CLA only allows to link one email address to a contributor. But I'm using multiple addresses and if I do a change here on GitHub directly, an address is picked that is not linked to my CLA account. I cannot update on the eclipse side because the address is used in other eclipse projects.

But for one liners I'm not really willing to clone, change, push if the Webinterface can do it more easily.

If you like to, you may integrate the changes on your behalf though.

dnav:
I hear you. But this is the Eclipse IP policy. My hands are tied.

Even doing the change myself would infringe your intellectual property...

dnav:
Addressed in #390

Issue #382, opened 2018-07-14
exception in observe_remove()


Kevin0626:
@davidgraeff , We met an exception in the LWM2M server application. When the problem happens, the client process was just terminated, and a obs request was sent and eventually timeout.

note: there was already obs on the object when the new request was sent. It was triggered by a protection to ensure the obseration is alway on - the software will send a new obs if no notified data recieved exceed the pmax period.

This mechnism may be questionable, however we don't expect any exception like this. We tried to find out the reason from the stack but no clue on the reason.

Thread #12 [lwm2m_main] 18058 [core: 0] (Suspended : Signal : SIGABRT:Aborted)	
	__GI_raise() at raise.c:54 0x7ffff7604428	
	__GI_abort() at abort.c:89 0x7ffff760602a	
	__libc_message() at libc_fatal.c:175 0x7ffff76467ea	
	malloc_printerr() at malloc.c:5,006 0x7ffff764f37a	
	_int_free() at malloc.c:3,867 0x7ffff764f37a	
	__GI___libc_free() at malloc.c:2,968 0x7ffff765353c	
	lwm2m_free() at platform.c:33 0x7fffeea58ed0	
	observe_remove() at observe.c:791 0x7fffeea50a4b	
	prv_obsRequestCallback() at observe.c:837 0x7fffeea50b3f	
	transaction_send() at transaction.c:418 0x7fffeea4d710	
	transaction_step() at transaction.c:443 0x7fffeea4d7a6	
	lwm2m_step() at liblwm2m.c:476 0x7fffeea4799b	
	thread_lwm2m() at lwm2mserver.c:160 0x7fffeea43410	
	ThreadWrapper() at 0x7ffff79b49fe	
	start_thread() at pthread_create.c:333 0x7ffff71b56ba	
	clone() at clone.S:109 0x7ffff76d641d	

=== prv_obsRequestCallback locals =========================
static void prv_obsRequestCallback(lwm2m_transaction_t * transacP,
                                   void * message)
{

    if (code != COAP_205_CONTENT)
    {
        observationP->callback(observationP->clientP->internalID,
                               &observationP->uri,
                               code,
                               LWM2M_CONTENT_TEXT, NULL, 0,
                               observationP->userData);
        observe_remove(observationP);
    }
locals dump:
=========
observationP	lwm2m_observation_t *	0x7fffd8000a70	
	next	struct _lwm2m_observation_ *	0x0	
	id	uint16_t	0	
[...]

PR #393, opened 2018-09-28
issue 382: observation memory free exception

Kevin0626:
This patch is created to fix the issue #382. The solution followed the proposal provided in the #382.

    Don't hold the pointer to client and observation object in the transaction object since many possible situations can happen during the transaction. for example, the client could reset, the user application could cancel the obs or start a new observation on the same client resource.

    Ensure in any case the user application can have a chance to handle cleanup for the pointer "UserData" passed in the lwm2m_observe() and lwm2m_observe_cancel() , through the callback function it also passed into the functions.


Kevin0626:
@dnav To sign the commit, I have closed previous PR and re-created this PR. Could you please check if it can be accepted?

PR #396, opened 2018-10-31
free memory on successful and failed observation read object.

maxvankessel:
**Problem** 
Data isn't freed after a failed readout on an observation request.

**Explanation**
After an observe option is send to the client, a data readout on the object is preformed.
The callback `target->readFunc` is been called (data is already allocated), if this fails, data isn't freed.
Only on a successful readout.

**Steps to reproduce** 
1. Create a resource which returns an error.
2. Observe this resource. 
3. Check `core/management.c` line  `208`

### DIFF
--- a/core/management.c
+++ b/core/management.c
@@ -234,8 +234,8 @@ uint8_t dm_handleRequest(lwm2m_context_t * contextP,
                             LOG_ARG("Observe Request[/%d/%d/%d]: %.*s\n", uriP->objectId, uriP->instanceId, uriP->resourceId, length, buffer);
                         }
                     }
-                    lwm2m_data_free(size, dataP);
                 }
+                lwm2m_data_free(size, dataP);
             }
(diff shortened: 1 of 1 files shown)


sbertin-telular:
It seems to me that object_readData should free the data when it returns an error rather than requiring every caller to handle it.

maxvankessel:
Should I change it? Or will it be fixed?

sbertin-telular:
Just my opinion. If you agree, go ahead and change it. If not, then it is up to the maintainer (@dnav) to merge or request changes.

dnav:
Hi,

Thanks for this.

For being able to merge your PR you need to sign the Eclipse CLA and sign your commits with the same email address. See: https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Foundation_Contributor_License_Agreement

Regards,

sbernard31:
This should be fixed in master since #435 was integrated.

PR #418, opened 2019-02-05
LWM2M 1.1 registration

sbertin-telular:
This updates the registration procedure for LWM2M 1.1. Several changes were needed to make this happen.

First, the build needed to be updated to supply which version of the LWM2M spec to build for.

The new data types for LWM2M 1.1 (Unsigned Integer and Corelnk) are added. This wasn't made specific to the LWM2M 1.1 build because it is only internal and not exposed through the protocol.

Transaction callbacks were changed to pass the context. It is only used here for the registration procedure, but may have additional uses.

The tests were updated to actually run all the tests and to test unsigned integer conversions. Existing number conversion tests were expanded and made to test conversion of the same numbers both ways.

The example client and bootstrap server had some of the new optional Server object resources added to facilitate testing.

Several minor bugs were fixed and minor changes made to reduce compiler warnings and improve logging output.


dnav:
Thank you for all this work. I'm still reviewing it.

sbertin-telular:
I'm having second thoughts about adding the Corelnk type. I don't think it would actually be used as a type for a resource. What do you think, should I revert cc55289?

I realize this is a large chunk of changes, but some of it didn't make sense unless put in context. I'll try to keep future changes smaller and more focused. I should have more ready to go in the next day or so, but they build on some of this so I'll wait for this pull request to complete before creating additional pull requests.

dnav:
Hi,

Why having second thoughts ? CoreLnk type is a possibility in 1.1 so it is better to be able to handle it in wakaama.

It looks fine for me.

PR #419, opened 2019-02-05
Improve floating point to text conversion

sbertin-telular:
Floating point conversion had a limited range that was supported. This improves the range but requires space in the buffer for the trailing null.


sbertin-telular:
I think there is a better way to do this, so I'm closing it and will rework it.

PR #431, opened 2019-02-21
[Core reorg] Organize source files per area.

dnav:
Signed-off-by: David Navarro <david.navarro@ioterop.com>


coap/coap.cmake
```diff
@@ -0,0 +1,16 @@
...
+    
+set(COAP_SOURCES 
+    ${COAP_SOURCES_DIR}/transaction.c
+    ${COAP_SOURCES_DIR}/block1.c
+    ${COAP_SOURCES_DIR}/er-coap-13/er-coap-13.c)
+
+
+# This will not work for multi project cmake generators like the Visual Studio Generator
+if(CMAKE_BUILD_TYPE MATCHES Debug)
+   set(WAKAAMA_DEFINITIONS ${WAKAAMA_DEFINITIONS} -DLWM2M_WITH_LOGS)
```
sbertin-telular: I don't think this should be here or in data.cmake, just wakaama.cmake. LWM2M_WITH_LOGS is only used in files under core and this modifies WAKAAMA_DEFINITIONS that is out of scope here.

core/internals.h
```diff
@@ -61,7 +61,7 @@
 #include <stddef.h>
 #include <stdbool.h>
 
-#include "er-coap-13/er-coap-13.h"
+#include "../coap/er-coap-13/er-coap-13.h"
```
sbertin-telular: coap was added to the include directories. I don't think it is necessary to add the ../coap part of the path here also.

sbertin-telular:
internals.h now has declarations for .c files that were moved to different directories. It should probably be broken up to keep the declarations near the definitions.

sbertin-telular:
This can be closed now. Replaced by #528 .

PR #432, opened 2019-02-22
Fix potential type size mismatch through pointers

### LINKED ISSUE #429: Typecasting causing undefined behavior
In er-coap-13.c there is the following code:
```
    x = &option_delta;
    do
    {
      if (*x==13)
      {
        *x += current_option[0];
        ++current_option;
      }
      else if (*x==14)
      {
        *x += 255;
        *x += current_option[0]<<8;
        ++current_option;
        *x += current_option[0];
        ++current_option;
      }
    }
    while (x!=(unsigned int *)&option_length && (x=(unsigned int *)&option_length));
```
Both `option_delta` and `x` are of type `unsigned int`, while `option_length` is of type `size_t`. In some systems these types are of different sizes, thus such code causes undefined behavior. My suggestion would be to change the type of `option_length` to `unsigned int`.

sbertin-telular:
size_t can be 64-bits while unsigned int is smaller. This causes issues
when option_length is accessed through the pointer x in
coap_parse_message(). The option length and delta are at most a 16-bit
unsigned integer plus 269 according to RFC 7252. Changing
option_number, option_delta, and option_length to uint32_t and x to
uint32_t * fixes #429. This also ensures sufficient size even if used
on a platform with smaller integers.

### DIFF
--- a/core/er-coap-13/er-coap-13.c
+++ b/core/er-coap-13/er-coap-13.c
@@ -617,10 +617,10 @@ coap_parse_message(void *packet, uint8_t *data, uint16_t data_len)
 {
   coap_packet_t *const coap_pkt = (coap_packet_t *) packet;
   uint8_t *current_option;
-  unsigned int option_number = 0;
-  unsigned int option_delta = 0;
-  size_t option_length = 0;
-  unsigned int *x;
+  uint32_t option_number = 0;
+  uint32_t option_delta = 0;
+  uint32_t option_length = 0;
+  uint32_t *x;
 
   /* Initialize packet */
   memset(coap_pkt, 0, sizeof(coap_packet_t));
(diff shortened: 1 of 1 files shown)


sbertin-telular:
@rettichschnidi please review.

sbertin-telular:
Rebased

core/er-coap-13/er-coap-13.c
```diff
@@ -617,10 +617,10 @@ coap_parse_message(void *packet, uint8_t *data, uint16_t data_len)
...
   coap_packet_t *const coap_pkt = (coap_packet_t *) packet;
   uint8_t *current_option;
-  unsigned int option_number = 0;
-  unsigned int option_delta = 0;
-  size_t option_length = 0;
-  unsigned int *x;
+  uint32_t option_number = 0;
+  uint32_t option_delta = 0;
+  uint32_t option_length = 0;
+  uint32_t *x;
```
rettichschnidi, line 623: The casts on line 699 should to be adjusted too.

Rest is looking good to me.

sbertin-telular:
Amended to fix casts

sbertin-telular:
Wait a minute, the casts shouldn't be necessary now.

sbertin-telular:
Amended to remove unnecessary type casts.

PR #433, opened 2019-02-28
Add files via upload

waterzhou:
align parameter method's type with coap protocol, it exceeds the define of coap_method_t already. its type should align with coap_init_message(..,.., method,..),


sbertin-telular:
[RFC8132](https://tools.ietf.org/html/rfc8132) adds FETCH, PATCH, and iPATCH methods. The coap_method_t enum does not contain those methods. This changes the method parameter of transaction_new from coap_method_t to uint8_t. I think it would be better to add those methods to the coap_method_t enum if they are needed.

sbertin-telular:
@waterzhou do you have any feedback on my comment? If not, I think we can close this PR without merging.

sbernard31:
I close this PR, but @waterzhou feel free to reopen it or add any comments if needed. :slightly_smiling_face:

PR #455, opened 2019-10-14
Fix DTLS connection_send in example code

tobbe303:
The number of bytes sent returned from send_data was interpreted as
error.

Signed-off-by: Tobias Gustafsson <tobias.gustafsson@inux.se>

### DIFF
--- a/examples/shared/dtlsconnection.c
+++ b/examples/shared/dtlsconnection.c
@@ -554,7 +554,7 @@ void connection_free(dtls_connection_t * connList)
 int connection_send(dtls_connection_t *connP, uint8_t * buffer, size_t length){
     if (connP->dtlsSession == NULL) {
         // no security
-        if ( 0 != send_data(connP, buffer, length)) {
+        if (0 >= send_data(connP, buffer, length)) {
             return -1 ;
         }
     } else {


examples/shared/dtlsconnection.c
```diff
@@ -554,7 +554,7 @@ void connection_free(dtls_connection_t * connList)
 int connection_send(dtls_connection_t *connP, uint8_t * buffer, size_t length){
     if (connP->dtlsSession == NULL) {
         // no security
-        if ( 0 != send_data(connP, buffer, length)) {
+        if (0 > send_data(connP, buffer, length)) {
```
sbertin-telular, line 557: This would treat 0 length as an error. I would suggest "0 >=" instead. Of course send_data doesn't pass a 0 length on to sendto, but that is a different issue.

sbernard31: @tobbe303, what do you think ? do you plan to adapt the code if you agree ?

@sbertin-telular, if we don't get any news in few days I will integrate your fix. (I mean using `>=`)

tobbe303: @sbernard31 Agree. I've updated the code as suggested.

sbernard31:
I cherrypicked your commit in master (8784fac655f7baaed3ddc8fa9430655c851bb1f4)

@tobbe303 thx for you contribution :pray: !

PR #463, opened 2020-02-20
Resolving base64 decoding bug

Jacob-Hinton:
Base64 decoding requires that adjacent variables from the encoded string are combined to generate the final correct binary data stream. The current implementation will result in every 3rd byte decoded into junk, this fix resolves the issue.


sbernard31:
Generally to accept your contribution we need you [to sign ECA](https://wiki.eclipse.org/Development_Resources/Contributing_via_Git#Eclipse_Contributor_Agreement).

As this PR is just about 1 character, I pretty sure there is not possible IP issue, so I just integrated using my name. (commit 178c1a7cb180ab56fa26e3de796a2dfbbadb92ce)

@Jacob-Hinton, thx for you contribution :pray: !

PR #461, opened 2020-01-14
Block transfer support

tuve:
Full block transfer support including on both client and server including:
- block1 registrations
- block1 write requests
- block2 read responses
- block2 observe responses
- block transfer retransmission on entity too large response

Signed-off-by: Tuve Nordius <tuve.nordius@husqvarnagroup.com>

PR #476, opened 2020-08-18
Block transfer support

LukasKarel:
fix mistakes of #461


LukasKarel:
on second inspection, this implementation does not seem to solve our problems.

PR #480, opened 2020-09-08
if the registration_getStatus function (registration status) returns …

ArtemTheVan:
If the registration_getStatus function (registration status) returns STATE_REG_FAILED and if bootstrap is not used switch contextP-> state to STATE_REGISTER_REQUIRED, not STATE_BOOTSTRAP_REQUIRED

### DIFF
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -442,8 +442,12 @@ next_step:
             break;
 
         case STATE_REG_FAILED:
+#ifdef LWM2M_BOOTSTRAP
             // TODO avoid infinite loop by checking the bootstrap info is different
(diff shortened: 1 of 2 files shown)


sbertin-telular:
I disagree with the changes to core/liblwm2m.c. This sets up for an infinite loop that may not be recoverable. The application needs to know about the failure so that it can attempt to deal with it. This may involve resetting a radio, setting up security objects with a different server/credentials, or simply retrying later.

sbernard31:
@sbertin-telular, should we close this issue or is there some part of this PR which makes sense ?

sbertin-telular:
I think we can close it. The other change prevents reporting and handling the failure of lwm2m_step. That also seems undesirable.

PR #485, opened 2020-09-30
fix dataArray size of client example's firware

tkerdonc:
Encountered when running the example on ARM, looks like the dataArray size is off by three in the client's firmware object, causing the client to segfault when one attempts to read the whole object.
Funnily enough, does not occur on x86.


sbernard31:
It seems this is already fixed in master : https://github.com/eclipse/wakaama/blob/9f12a00790ba64c0570f8d2178bf165ba453b146/examples/client/object_firmware.c#L92

@tkerdonc thx for the contribution anyway :wink:

Issue #352, opened 2018-01-31
build wakaama with libcoap backend


ponyatov:
Is it possible to strip out all COAP specific code from Wakaama sources to use external libcoap?
Looking at the code I see COAP specific code all over the sources.

I'm going to use a tiny device with RT5350F, and it will be good to have variants to use full LWM2M or minimal COAP client mode depending on the current task.

sbernard31:
Do you find a solution for this ? 
@sbertin-telular is it something we want to address ?

sbertin-telular:
I see the value in being able to build with different CoAP libraries, but I think Wakaama is closely tied to the CoAP code right now. I think we should leave it open for future work.

sbernard31:
Maybe we change label to `enhancement` ? (or anything else you prefer)

sbertin-telular:
I think enhancement is good. Go ahead and change it.

sbernard31:
(I just get that you still not have right to change label :disappointed:)

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #492, opened 2020-11-04
Fix reference to Erbium's CoAP engine

rettichschnidi:
* The original website is no longer available, therefore linking to latest version mirrored on archive.org.
* Even if the changes where just small in the beginning, nowadays there is a big delta. Calling the modifications "slightly" is misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Developers mailing list: https://dev.eclipse.org/mailman/listinfo/wakaama-dev
 
     -+- core                   (the LWM2M engine)
      |    |
-     |    +- er-coap-13        (Slightly modified Erbium's CoAP engine from
+     |    +- er-coap-13        (Modified Erbium's CoAP engine from
      |                          https://web.archive.org/web/20180316172739/http://people.inf.ethz.ch/mkovatsc/erbium.php)
      |
      +- tests                  (test cases)
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Developers mailing list: https://dev.eclipse.org/mailman/listinfo/wakaama-dev
     -+- core                   (the LWM2M engine)
      |    |
      |    +- er-coap-13        (Slightly modified Erbium's CoAP engine from
-     |                          http://people.inf.ethz.ch/mkovatsc/erbium.php
+     |                          https://web.archive.org/web/20180316172739/http://people.inf.ethz.ch/mkovatsc/erbium.php)
      |
      +- tests                  (test cases)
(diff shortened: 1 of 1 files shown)


sbertin-telular (approved):

README.md
```diff
@@ -8,7 +8,7 @@ Developers mailing list: https://dev.eclipse.org/mailman/listinfo/wakaama-dev
     -+- core                   (the LWM2M engine)
      |    |
      |    +- er-coap-13        (Slightly modified Erbium's CoAP engine from
```
sbernard31, line 10: I just looked at the Contiki repository : https://github.com/contiki-os/contiki/tree/master/apps/er-coap

At first sight, the code seems more than "slightly modified" (I confess that my judgement is only based on number of files and files name, so it could be totally wrong  :sweat_smile:)

**Naive question**, this does not make sense to synchronize those 2 projects ? (I mean erbium in contiki and erbium in wakaama ?)

sbertin-telular: At this point it might be a similar level of effort as #352 to update to the latest erbium code, but I haven't looked at it in detail. Why don't we create an issue to track that.

sbernard31: > Why don't we create an issue to track that.

Yep, if you think it worth to look at,  that makes sense :)

sbernard31: About "slightly modified" we let it like this even if this is maybe more than that ?
(This is not so important I suppose)

rettichschnidi: Just checked out the [initial commit](https://github.com/eclipse/wakaama/commit/b1691d28c56b28ca1c317e79f365c3c282f8135e) of Wakaama, tried to find the matching sources in the [Contiki](https://github.com/contiki-os/contiki/) code base/history - I failed. It seems to me that Wakaama's copy of the code got changed even before the first commit.

I sent an email to the creator of the first commit, asking if there might still be a previous repository lying around somewhere. I guess this could provide us a conclusive answer.

rettichschnidi: > About "slightly modified" we let it like this even if this is maybe more than that ?
> (This is not so important I suppose)

I removed the "slightly" in the statement - even if the changes introduced by Wakaama turn out to be small, the changes which in the Contiki project since leave us with quite a big delta, rendering the phrase misleading.

sbernard31: @rettichschnidi before to be hosted by eclipse foundation, the project was host here : https://github.com/intel/liblwm2m

rettichschnidi: @sbernard31 Created #495 using this information.

PR #512, opened 2020-12-16
Improve coding style

### LINKED ISSUE #505: uriP pointer is used as boolean in uri_toString() in the uri.c file. Line 292 and 338.

qleisan:
Minor style improvement in uri.c

Fixes #505

### DIFF
--- a/core/uri.c
+++ b/core/uri.c
@@ -289,7 +289,7 @@ int uri_toString(const lwm2m_uri_t * uriP,
     LOG_ARG("bufferLen: %u", bufferLen);
     LOG_URI(uriP);
 
-    if (uriP && LWM2M_URI_IS_SET_OBJECT(uriP))
+    if (NULL != uriP && LWM2M_URI_IS_SET_OBJECT(uriP))
     {
         int res;
         depth = URI_DEPTH_OBJECT;
@@ -335,7 +335,7 @@ int uri_toString(const lwm2m_uri_t * uriP,
             }
         }
     }
-    else if(uriP)
+    else if(NULL != uriP)
     {
         if (head >= bufferLen - 1) return -1;
         buffer[head++] = '/';


core/uri.c
```diff
@@ -335,7 +335,7 @@ int uri_toString(const lwm2m_uri_t * uriP,
             }
         }
     }
-    else if(uriP)
+    else if(NULL != uriP)
```
sbernard31, line 338: Is there a reason to write `NULL != uriP` instead of `uriP != NULL` ?
I feel the second a bit more "natural" ?

qleisan: Personally I prefer the readability of (variable == NULL) but (NULL == variable) is often considered safer since if you mistype "==" as "=" the latter causes compilation error.
However "follow the existing coding style" usually is considered most important (I see plenty of both). Shall I change? (need to solve "eca" problem with husqvarna e-mail anyway)

sbernard31: @qleisan thx for the explanation.

I think both solution makes sense.
I guess the team should choose the way which should be applied for this project.
This makes me thing that this could be the beginning of a kind of code styling & rules for this project : #513.

qleisan:
Ok to merge this PR as is? It solves the issue, if we set an official coding standard for "value == var or var == value" there are many instances that should be updated.

sbertin-telular:
I'm not entirely sure there is an issue to solve. What does adding the comparison to NULL gain?

sbernard31:
Probably not an issue, just another code styling the team need to agree on :grin:  (https://stackoverflow.com/questions/3825668/checking-for-null-pointer-in-c-c)

qleisan:
I also think that adding a comparison with NULL doesn't add much value, especially when multiple comparisons are done on the same line with "and" (&&) operator, makes it harder to read. When searching the webb however I got the impression that most coding standards favored doing the comparison and it seems to be the style (mostly) used in this project.

sbertin-telular:
If it is mostly used already, then I'm fine with it.

core/uri.c
rettichschnidi: Regarding the safety: I think this is no longer relevant as we have compiler warnings for this nowadays.

sbernard31:
So I understand that : 
1/ you decide to do comparison with NULL because this is already mostly used in the project ? correct ? (I don't know if this should be discuss at  #513)
2/ for the order `var == value` or `value == var`, reading comment I understand that everybody feel the first one more readable and finally there is no good reason to do the second one ? or I missed something ?

Sorry if this looks like  "nitpicking", but as I feel this PR is just fixing code styling, I think there is no urgency and before change this it is probably better to decide general rules to apply :slightly_smiling_face:.

qleisan:
I felt this was such a minor fix that it would be nice to close the issue now. It has consumed so much attention already. We have plenty of "var == value or value == var" and similar to fix once we decide on preferred coding style. The "bigger" question about doing a comparison with NULL (boolean expression) is addressed.  Nice to close issues that clutter the view.. 
I'll close this now and we can make a more general PR later addressing #505 and and many other instances of code that should be aligned with preferred coding style (TBD)

PR #520, opened 2021-01-12
Enable SENML_JSON support in lightclient

qleisan:
Lightclient example now works out-of-the-box with wakaama server example.

Wakaama client / server examples should work together after a "vanilla build" to provide a good user experience.

### DIFF
--- a/examples/lightclient/CMakeLists.txt
+++ b/examples/lightclient/CMakeLists.txt
@@ -10,6 +10,10 @@ include(${CMAKE_CURRENT_LIST_DIR}/../../core/wakaama.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/../shared/shared.cmake)
 
 add_definitions(-DLWM2M_CLIENT_MODE)
+if(LWM2M_VERSION VERSION_GREATER "1.0")
+    add_definitions(-DLWM2M_SUPPORT_SENML_JSON)
+endif()
+
 add_definitions(${SHARED_DEFINITIONS} ${WAKAAMA_DEFINITIONS})
 
 include_directories (${WAKAAMA_SOURCES_DIR} ${SHARED_INCLUDE_DIRS})


sbertin-telular:
The server should support all formats. I think the JSON formats were added to the other client to support command line enhancements.

This makes me wonder what format the lightclient was using before this change. I lost everything on my hard disk recently and haven't gotten set back up for building Wakaama to check. If it supported no formats, we should probably add something to liblwm2m.h to give a reasonable default for people writing their own client.

sbertin-telular:
> This makes me wonder what format the lightclient was using before this change. I lost everything on my hard disk recently and haven't gotten set back up for building Wakaama to check. If it supported no formats, we should probably add something to liblwm2m.h to give a reasonable default for people writing their own client.

I checked. The lightclient only supported plain text and probably opaque. Nothing that could represent any structured data.

qleisan:
TLV was supported until `ab776be "Make TLV format optional."`. The 1.1 spec states 

> "LwM2M Clients supporting LwM2M TS 1.1 or newer SHOULD NOT use the legacy TLV or JSON formats" 

Although the lightclient has no CLI we still need to support a valid format for the payload, right? (any finer points about "structured data" that needs to be highlighted?)

The logic for TLV support or not is handled differently (header file logic) from SENML/JSON (cmake file logic). I used the pattern from the regular client to adapt the lightclient. Any thoughts on this?

examples/lightclient/CMakeLists.txt
```diff
@@ -10,6 +10,12 @@ include(${CMAKE_CURRENT_LIST_DIR}/../../core/wakaama.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/../shared/shared.cmake)
 
 add_definitions(-DLWM2M_CLIENT_MODE)
+if(LWM2M_VERSION VERSION_GREATER "1.0")
+    add_definitions(-DLWM2M_SUPPORT_SENML_JSON)
+else()
+    add_definitions(-DLWM2M_SUPPORT_JSON)
```
sbertin-telular, line 16: TLV will be supported by default for LWM2M 1.0. For a lightweight client I don't think we need to add JSON also.

sbertin-telular:
For LWM2M 1.1 there is no clear default once SENML CBOR support is added. I'm good with having it in the cmake file. This solves it for lightclient and I created #523 for the general case.

qleisan:
@sbernard31 @sbertin-telular Is this good to merge now? (or shall I remove LWM2M_SUPPORT_JSON for the 1.0 case since the lightclient doesn't have a CLI)

sbernard31:
(I let you decide about this, I merge it as soon as the PR is approved :slightly_smiling_face:)

sbertin-telular:
> (or shall I remove LWM2M_SUPPORT_JSON for the 1.0 case since the lightclient doesn't have a CLI)

I'd like to see it removed. It is supposed to be a lightweight example. Adding unnecessary features is counter to that.

sbernard31:
(@qleisan, if you adapt the PR, could you rebase+push force)

qleisan:
Agree! New version pushed

sbertin-telular (approved):

sbertin-telular (approved):

PR #522, opened 2021-01-18
Fix passing NULL pointer to printf as string.

CiaranWoodward:
Doing this is undefined behaviour, and causes crashes with certain
compilers/C libraries. The LOG_ARG macro is defined to function as
printf.

This was causing issues when running wakaama with logs enabled.

I went through all of the LOG_ARG calls, and wrapped any string args
which could be NULL in a helper macro which swaps them for an empty
string.

Signed-off-by: Ciaran Woodward <c.woodward@cascoda.com>

### DIFF
--- a/core/data.c
+++ b/core/data.c
@@ -211,7 +211,7 @@ void lwm2m_data_encode_string(const char * string,
     size_t len;
     int res;
 
-    LOG_ARG("\"%s\"", string);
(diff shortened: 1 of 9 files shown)


sbertin-telular (approved):
Thank you for this contribution. I like that it only adds the null pointer check when logging is enabled.

Issue #355, opened 2018-02-06
Lifetime update and registration update


fdur:
According to OMA-TS-LightweightM2M-V1_0_1-20170704-A, in §5.3.2 (Update):

>When any of the parameters listed in Table 9 changes, the LwM2M Client MUST send an “Update” operation to the LwM2M Server. This “Update” operation MUST contain only the parameters listed in Table 9 which have changed compared to the last registration parameters sent to the LwM2M Server.
>
Parameter |	Required
--------------| ---------------
Lifetime | No
Binding Mode | No
SMS Number | No
Objects and Object Instances | No

Table 9: Update parameter

and

> 1)	Extending the lifetime of a registration
In this case the Client sends a Registration Update with no parameters.
Figure 10 shows an example exchange where the Client sends a Registration Update message that only refreshes the registration, i.e., the message does not contain any parameters. With the second Registration Update the Client changes the lifetime field to 6000 (seconds) and hence the lt parameter is included in the message.

When the DM server updates the lifetime value, the registration update message needs to be sent to this server without the new value
This registration message is not sent

cdarisi3:
@fdur : Hi, Do you know any fix on this issue. Appreciate your help

sbernard31:
> When the DM server updates the lifetime value, the registration update message needs to be sent to this server without the new value

you mean : "the registration update message needs to be sent to this server ~without~ with the new value" ?

fdur:
> you mean : "the registration update message needs to be sent to this server ~without~ with the new value" ?
Yes, sorry for the misunderstanding :)

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #529, opened 2021-02-01
Reflect changes to the server object into the core stack

### LINKED ISSUE #209: An update of the lifetime parameter of a server object instance has no effect
An update of the lifetime parameter of a server object instance has no effect on the actual lifetime used for registration update. 

At the start of the client, a server object is created using the values of the security object instance and server object instance provided to the client. This server object gets the lifetime value by copying the value from the server object instance ( prv_getMandatoryInfo(...) ). 

This newly created server object, which has a copy of the lifetime value of the server object insatnce, is used when deciding if we need a registration update. So if we update the server object instance, it does not affect the actual lifetime value used by the client.

CiaranWoodward:
Updating the server object didn't have an impact after connecting to the server, so it was impossible to change the lifetime/binding after bootstrap. This modification catches writes to the lifetime object and immediately updates the internal representation of the server.

I'm not overly familiar with wakaama, so please check for any blatant errors! It seems a bit strange to have to extract the values from the server object using such a heavy-handed method, but it seems the stack is designed to keep the objects very detached from the wakaama internals. I was otherwise considering making the server object itself (object_server.c) modify the internal stack state (by directly accessing the server list), however the server object seems otherwise isolated from the exact instance it is attached to, so I tried to stay in the spirit of that.

Resolves #209
Related but unresolved: #355

Signed-off-by: Ciaran Woodward <c.woodward@cascoda.com>

### DIFF
--- a/core/objects.c
+++ b/core/objects.c
@@ -59,6 +59,52 @@
 #include <string.h>
 #include <stdio.h>
 
+static int prv_getMandatoryInfo(lwm2m_object_t * objectP,
+                                uint16_t instanceID,
+                                lwm2m_server_t * targetP)
+{
+    lwm2m_data_t * dataP;
+    int size;
+    int64_t value;
+
+    size = 2;
+    dataP = lwm2m_data_new(size);
+    if (dataP == NULL) return -1;
+    dataP[0].id = LWM2M_SERVER_LIFETIME_ID;
+    dataP[1].id = LWM2M_SERVER_BINDING_ID;
+
+    if (objectP->readFunc(instanceID, &size, &dataP, objectP) != COAP_205_CONTENT)
+    {
+        lwm2m_data_free(size, dataP);
+        return -1;
+    }
+
+    if (0 == lwm2m_data_decode_int(dataP, &value)
+        || value < 0 || value >0xFFFFFFFF)             // This is an implementation limit
+    {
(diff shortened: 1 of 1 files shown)


core/objects.c
```diff
@@ -262,6 +308,57 @@ uint8_t object_read(lwm2m_context_t * contextP,
...
+
+    if (serverObjectP->readFunc(instanceId, &size, &dataP, serverObjectP) != COAP_205_CONTENT)
+    {
+        lwm2m_data_free(size, dataP);
+        return 0;
+    }
+
+    if (1 != lwm2m_data_decode_int(dataP, &value))
+    {
+        value = -1;
```
sbertin-telular, line 331: This will return 65535. Should be 0.

core/objects.c
```diff
@@ -262,6 +308,57 @@ uint8_t object_read(lwm2m_context_t * contextP,
...
+    }
+    lwm2m_data_free(size, dataP);
+
+    if(value > UINT16_MAX)
```
sbertin-telular, line 335: Should also check for negative values.

CiaranWoodward:
@sbertin-telular Thanks for the review, I have resolved your comments & force pushed

sbertin-telular (approved):

sbernard31:
@CiaranWoodward  thanks for the contribution :pray:  !

Issue #513, opened 2020-12-16
Define code styling rules


sbernard31:
Maybe it would be good to define some rules about code styling ? This could help to have a more consistent code base ?

The rules could be written in a wiki page (e.g. https://github.com/eclipse/leshan/wiki/Code-&-design-guidelines)

sbernard31:
E.g of question : 
Should we write `value == var` or `var == value` (see https://github.com/eclipse/wakaama/pull/512#discussion_r544283993 for more details) ?

@sbertin-telular, @tuve, @qleisan, @rettichschnidi ?

rettichschnidi:
I'd split this into two parts:
* Code formatting: using clang-format (not part of this issue)
* Everything else (including the example from above): Lets find something which is a) close to the predominantly used style in this code base (is there one?) and b) is already existing

sbertin-telular:
As far as `value == var` or `var == value`, I find `var == value` slightly easier to read and write as it aligns with the way I think. I have been working on a project recently with `value == var` as the standard, so I'm getting used to it, but it does still take a bit of extra time to write or read it that way.

sbernard31:
(I'm not sure be I think you have right to create wiki page if needed)

sbernard31:
Add a new code style question  : `if (ptr)` VS `if(ptr != NULL)` (from #512) 
I will add all this kind of question in the issue description :point_up:

qleisan:
Decide coding style for when to check pointers
https://github.com/eclipse/wakaama/issues/514#issuecomment-761041820
https://github.com/eclipse/wakaama/issues/506#issue-752058734

sbernard31:
Question about usage of  `assert` : https://github.com/eclipse/wakaama/pull/531

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #531, opened 2021-02-03
Remove incomplete error state in transaction_new

CiaranWoodward:
Discovered what looked like a potential memory leak in the 'error' state of transaction_new. However it wasn't actually possible to hit this state without a programmer error (utils_intToText will never fail with a length 6 buffer and a uint16_t int)... but the state did exist and didn't properly clear up. One option was to correct it, so that it cleaned up properly:

```C
error:
    LOG("Exiting on failure");
    if(transacP->message) 
    {
        coap_free_header(transacP->message);
        lwm2m_free(transacP->message);
    }
    lwm2m_free(transacP);
    return NULL;
```

But I figured it was cleaner to just remove the unused & incomplete code.

### DIFF
--- a/coap/transaction.c
+++ b/coap/transaction.c
@@ -240,6 +240,11 @@ lwm2m_transaction_t * transaction_new(void * sessionH,
 
 error:
     LOG("Exiting on failure");
+    if(transacP->message)
+    {
+        coap_free_header(transacP->message);
+        lwm2m_free(transacP->message);
+    }
     lwm2m_free(transacP);
     return NULL;
 }


sbertin-telular:
I'm torn between approving of removing code paths that can't be executed and disapproving of the removal of error checking that could catch future problems. I'm leaning toward approval. Any other opinions?

sbernard31:
Classic dilemma. 

I guess this is a bet based on : 
- cost of checking
- risk prevented by this check

My 2 cts in this case : I feel that cost checks is very low but in a meantime I don't to see which kind of risk it could concretely prevent.(but maybe I just missed it)
So I would rather go for removing. (Clearly not strong opinion)

sbernard31:
@qleisan  any opinion about this ?

qleisan:
The potential malloc runtime error is handled and that is great. Potential runtime errors should always be checked and handled. 
Removal of the check for `result == 0` is questionable I think. Programming errors cannot / should not always be checked for but in this case it is a) a pattern used on many other places when the function is called b) a call to a function that explicitly returns 0 in several places. If the `result == 0` check is removed I think a comment should be added to help others feel good about this when reading the code, It does get cleaner without the check...

sbernard31:
I'm not sure we helped @CiaranWoodward to know what he should do :sweat_smile: 

My understanding :  I feel that the team is not totally comfortable with the idea of removing the check
But  this sounds clearly not to be a strong opinion.

So @CiaranWoodward, I guess this is up to you to decide what to do : 
1. either keep the error check
2. or remove it (but in this case a comment would be appreciated)

@sbertin-telular, @qleisan  do not hesitate to correct me if I misunderstood the situation.

CiaranWoodward:
Hi, apologies for being pretty silent here. I'll take another look and see if I can find a more graceful fix - can always fall back to the snippet I posted in the initial comment.

What is this project's stance on `assert``s? I noticed that there weren't any in the codebase, but they can be a good way to catch coding errors. Any thoughts?

CiaranWoodward:
I looked into it and seeing as it isn't that big of a deal for code size or stability, I reverted 25e495b5e76b1eec6b366e34938dc3a5abebfa04 and replaced it with the snipped from the comment. It does still seem out of place to me to catch an error that can never happen, but can chalk it up to different style. It may be worth looking into solutions like ``assert``, that can signal programmer intent and just crash the program upon error.

The new commit has been pushed - hopefully that solves the issue!

tuve (approved):

sbernard31:
@CiaranWoodward thx for your contribution :pray: 

@tuve, @qleisan , @sbertin-telular  do not hesitate to share your point of view about `assert` usage. (Just in case I added a comment about this in #513)

PR #532, opened 2021-02-03
Add received message size check

### LINKED ISSUE #509: If DATA is more than 1003 bytes only 1003 bytes are sent without any error message.
While testing #461

server: write 0 /31024/10/5 DATA
client: dump /31024/10

If DATA is more than 1003 bytes only 1003 bytes are sent without any error message. Possibly command line input limitation.

qleisan:
Received packets >= MAX_PACKET_SIZE are dropped. MAX_PACKET_SIZE increased to 2048

Resolves #509

### DIFF
--- a/examples/client/lwm2mclient.c
+++ b/examples/client/lwm2mclient.c
@@ -80,7 +80,7 @@
 #include <errno.h>
 #include <signal.h>
 
-#define MAX_PACKET_SIZE 1024
+#define MAX_PACKET_SIZE 2048
 #define DEFAULT_SERVER_IPV6 "[::1]"
 #define DEFAULT_SERVER_IPV4 "127.0.0.1"
 
@@ -1297,6 +1297,10 @@ int main(int argc, char *argv[])
                 {
                     fprintf(stderr, "Error in recvfrom(): %d %s\r\n", errno, strerror(errno));
                 }
+                else if (numBytes >= MAX_PACKET_SIZE) 
+                {
+                    fprintf(stderr, "Received packet >= MAX_PACKET_SIZE\r\n");
+                } 
                 else if (0 < numBytes)
                 {
                     char s[INET6_ADDRSTRLEN];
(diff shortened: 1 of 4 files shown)


examples/client/lwm2mclient.c
```diff
@@ -1297,7 +1297,11 @@ int main(int argc, char *argv[])
                 {
                     fprintf(stderr, "Error in recvfrom(): %d %s\r\n", errno, strerror(errno));
                 }
-                else if (0 < numBytes)
+                else if (numBytes >= MAX_PACKET_SIZE) 
+                {
+                    fprintf(stderr, "Received packet >= MAX_PACKET_SIZE\r\n");
+                } 
+                else
```
sbertin-telular, line 1304: 0 length packets are no longer excluded from processing. Did you check that they are handled correctly?

examples/lightclient/lightclient.c
```diff
@@ -514,7 +514,11 @@ int main(int argc, char *argv[])
                 {
                     fprintf(stderr, "Error in recvfrom(): %d %s\r\n", errno, strerror(errno));
                 }
-                else if (0 < numBytes)
+                else if (numBytes >= MAX_PACKET_SIZE) 
+                {
+                    fprintf(stderr, "Received packet >= MAX_PACKET_SIZE\r\n");
+                } 
+                else
```
sbertin-telular, line 521: 0 length packets are now processed. Are they handled correctly?

qleisan: this was not intended will push an update

examples/client/lwm2mclient.c
qleisan: this was not intended will push an update

qleisan:
Legacy server implementation processes packets of length 0 (as opposed to clients) @sbertin-telular - do you know why?

sbertin-telular:
I don't know why length 0 is processed by the server and not the client. I've never looked at it. I just noticed it when reviewing your changes. A 0 length packet could be sent as a form of keep-alive, but I think the example programs should just ignore it.

sbertin-telular (approved):

Issue #440, opened 2019-05-09
Incorrect CoAP result on security object action made by the DM server


fdur:
According to LwM2M 1.0.2 specification:
> The LwM2M Client MUST reject with an “Unauthorized” response code any LwM2M Server operation on the Security Object (ID: 0).

In current source code, "Not Found" is returned

sbernard31:
Not a big bug but probably better to fix it.
This is also true for  [LWM2M 1.1.](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#6-3-0-63-Device-Management-and-Service-Enablement-Interface)

qleisan:
@fdur thanks for reporting this bug! @sbernard31 please close once #533 is merged.

PR #533, opened 2021-02-04
Fix for issue #440: Crash with client in DTLS when "quit" is usedInco…

qleisan:
…rrect CoAP result on security object action made by the DM server

Signed-off-by: Frederic DUR <fdur@sierrawireless.com>

Fix for issue #440 by Frederic DUR (all credits intact)

### DIFF
--- a/core/management.c
+++ b/core/management.c
@@ -183,7 +183,7 @@ uint8_t dm_handleRequest(lwm2m_context_t * contextP,
 
     if (uriP->objectId == LWM2M_SECURITY_OBJECT_ID)
     {
-        return COAP_404_NOT_FOUND;
+        return COAP_401_UNAUTHORIZED;
     }
 
     if (serverP->status != STATE_REGISTERED


sbernard31:
I don't get the : 
> Signed-off-by: Frederic DUR fdur@sierrawireless.com
and the 
> Frederic Dur - Please refer to git log

Do you cherry pick code from somewhere or I missed something ?

qleisan:
Yes created PR from #440 giving cred to user

core/management.c
```diff
@@ -17,7 +17,8 @@
  *    Bosch Software Innovations GmbH - Please refer to git log
  *    Pascal Rieux - Please refer to git log
  *    Tuve Nordius, Husqvarna Group - Please refer to git log
- *    
+ *    Frederic Dur - Please refer to git log
```
sbernard31, line 20: Probably not a big deal, but I guess the author here should rater be the author of the code. 

My concern is about potential IP issue. (even if probability is very low here)
In particular because  "There is no valid ECA on file for fdur@sierrawireless.com." (checked with ECA Validation Tool of eclipse)

(Out of topic but to be honest I'm not fan of this kind of log in the header especially when we have tool like git history. Just to let you know this is not mandatory for eclipse project see [Tracking contributors in Git vs file header comment](https://bugs.eclipse.org/bugs/show_bug.cgi?id=435933#c18) but project is free to use it if wanted)

qleisan: In this case I reasoned that the change was trivial once the problem was pointed out and it was nicer to give some credit than not to, however I see the general problem. And I agree about the header comments.
Correct me if I am wrong but pointing out a problem doesn't give anyone any ownership of the solution. If the solution is obvious and trivial there is no IP either - anyone can come to the same result.
I suggest we close this PR (might as well close #440 also). I can submit the obvious fix in a new PR.
(making an effort to reduce the number of open TRs)

sbernard31: As I said my 'little' concern is just about this line in the header.

**I'm pretty sure there is no issue** but if there is that could be impacting, like e.g. we need to retrieve Frederic Dur to get a given agreement about license changing or something like this. 

I just say that maybe this is safer to avoid this and so just thanks to the contributor for reporting the issue.

But again I'm not sure of all of this. And I just give you some concern. Feel free to let it like this or ask to other member of the team.

If you finally want to adapt the PR, just push force this one rather than create a new one.

qleisan: Force pushed a brand new commit - 100% handcrafted by me (21 characters of hard work 😂). Please merge!

sbernard31: :grin:

PR #539, opened 2021-02-11
Support the preferred content format for Bootstrap-Request

sbertin-telular:

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,9 +137,6 @@ If DTLS feature enable:
 To launch a bootstrap session:
 ``./lwm2mclient -b``
 
-***Note:***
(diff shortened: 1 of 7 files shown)


qleisan:
(Started to review. Will resume later today)

qleisan (approved):

tuve (approved):
I agree with @qleisan that the README should be updated as well. But codewise i think this is fine.

sbertin-telular:
Thanks for the review. This is the problem integrating year+ old changes with new changes, something always gets missed. I'll update.

sbertin-telular:
I've updated the README.md and rebased.

qleisan (approved):

PR #543, opened 2021-02-22
Improve coding style

### LINKED ISSUE #499: prv_parseItem() in json.c use string as a boolean
**string is no boolean, do not use negation**

static int prv_parseItem(const uint8_t * buffer,
                        size_t bufferLen,
                         _record_t * recordP)
{
 .
 .

case _TYPE_STRING:
        if (0 != recordP->valueLen)
        {
            size_t stringLen;
            uint8_t *string = (uint8_t *)lwm2m_malloc(recordP->valueLen);
            **if (!string) return false;**
            stringLen = json_unescapeString(string, recordP->value, recordP->valueLen);
            if (stringLen)
            {
                lwm2m_data_encode_nstring((char *)string, stringLen, targetP);
                lwm2m_free(string);
            }

qleisan:
Make explicit comparison with NULL to address reported issue #499. Not a bug

Closes #499

### DIFF
--- a/data/json.c
+++ b/data/json.c
@@ -276,7 +276,7 @@ static bool prv_convertValue(_record_t * recordP,
         {
             size_t stringLen;
             uint8_t *string = (uint8_t *)lwm2m_malloc(recordP->valueLen);
-            if (!string) return false;
+            if (string == NULL) return false;
             stringLen = json_unescapeString(string, recordP->value, recordP->valueLen);
             if (stringLen)
             {


qleisan:
(forgot to sign off commit first time)

data/json.c
```diff
@@ -276,7 +276,7 @@ static bool prv_convertValue(_record_t * recordP,
         {
             size_t stringLen;
             uint8_t *string = (uint8_t *)lwm2m_malloc(recordP->valueLen);
-            if (!string) return false;
+            if (string == NULL) return false;
             stringLen = json_unescapeString(string, recordP->value, recordP->valueLen);
             if (stringLen)
```
sbertin-telular, line 281: If a NULL comparison for string is added, shouldn't there also be a comparison to 0 for stringLen?

qleisan: No check is really added, only rewritten `if(!string)` <=> `if (string == NULL)` to accommodate #499 concern about "using string as a boolean". I felt it was right to align to the dominating coding style of comparing with NULL.

I see your point but the coding style in the code is more explicit when checking pointers, I believe. Your input on #513 is appreciated

sbertin-telular (approved):

tuve (approved):

PR #547, opened 2021-03-01
Support builds for LWM2M 1.1.1

sbertin-telular:


qleisan (approved):
LGTM

sbernard31:
Even of this is not so clear, my understanding is that LWM2M Version used in the protocol is just composed of 2 number : `major.minor`. (See [LwM2M-Server-and-LwM2M-Client-Versions](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#5-2-1-0-521-LwM2M-Server-and-LwM2M-Client-Versions))
Knowing that `patch` number is only used for bug fixes and LWM2M version declared by client and server is used to ensure interoperability, I'm not sure sharing the bug fix version of the specification make sense :thinking:. 

For [object version](http://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#7-2-2-0-722-Object-Version-format) this is more clear. 

I will try to get more answer from OMA. 

@sbertin-telular is it something you need to resolve a specific problem or this is just your understanding of the specification ?

sbertin-telular:
There are minor differences between 1.1 and 1.1.1. I have allowed building either way in some of the changes I have upcoming. See https://github.com/sbertin-telular/wakaama/blob/84b9aa7663ec9e66b1666abd78b31151fcba40e3/core/bootstrap.c#L192 for an example. If we want to leave it as just 1.1 and have that include 1.1.1 changes, I'm OK with that. This is not to solve a specific problem, just to allow selection of which version of the specification to follow.

sbernard31:
I think I maybe didn't get the usage of `LWM2M_VERSION` of this PR. I was thinking this was about server lwm2m version "negotiation" at registration time. :grimacing: 
But this sounds not exactly this, I mean this more an internal stuff for wakaama, not exactly the "lwm2m version" sent or received at registration time. 
If I finally well understand it, this is not really a question for the OMA. 

> If we want to leave it as just 1.1 and have that include 1.1.1 changes, I'm OK with that. 

I have no strong opinion but maintaining several version of the spec could maybe be challenging.
So maybe targeting last version of each minor version is enough. (1.0 => 1.0.2, 1.1 => 1.1.1)
@qleisan, @tuve any opinions ?

qleisan:
Definitely in favor of

> targeting last version of each minor version is enough. (1.0 => 1.0.2, 1.1 => 1.1.1)

Unless there is a real need to differentiate.

Issue #536, opened 2021-02-09
Endless loop when client responds with large payload without block transfer


mlasch:
_Test scenario:_ A wakaama server issues a READ command to a client. 
When the client replies with a payload bigger than `REST_MAX_CHUNK_SIZE`, the wakaama server responds with "4.13 Entity too large" **and** advertises its configured block size with a new GET request and block2 option set. If the client does not handle the block2 GET correctly and again send a too large payload this results in an endless loop.
Ideas on how this can be handled, maybe a max retry count on the server side? I tested with wakaama server and zephyr lwm2m client.

sbernard31:
>  A wakaama server issues a READ command to a client.
> When the client replies with a payload bigger than REST_MAX_CHUNK_SIZE, the wakaama server responds with "4.13 Entity too large"

We should not answer to a response with a response. You can not send a 4.13 in response to a 2.05.
I mean 4.13 is [client response code](https://tools.ietf.org/html/rfc7252#section-5.9.2).
So there is maybe a first issue here. (or maybe I don't get you?)

> If the client does not handle the block2 GET correctly and again send a too large payload this results in an endless loop.

It sounds there is another issue here.

I guess this could be resolved like this.
- IF server receive a too big payload to a request sent **without block2** option THEN it could retry with block2 option.
- IF server receive a too big payload to a request sent **with block2** option THEN it should consider there is nothing to do and so report the error to application layer.

I guess this should avoid the infinite loop ? :thinking:

mlasch:
This screenshot illustrates a read command which shows the server behavior. (`REST_MAX_CHUNK_SIZE` is set to 512 on the server)
![zephyr_register_read_too_large](https://user-images.githubusercontent.com/17436433/107378788-989a3080-6aec-11eb-88f3-bc143ff39791.png)
This means the server should ignore the (too big) 2.05 response and only issue a new GET request with block2 set to `REST_MAX_CHUNK_SIZE`?

About your suggestions:
1. I agree, the server should indicate its block size with a second GET request.
2. If the client sends a block2 with a too big payload the server can still indicate its max. payload size with the block2 option. And hope the client can handle it correctly.

However after signaling its block size once, the server should stop trying and report an error.

sbernard31:
> This screenshot illustrates a read command which shows the server behavior.

Generally this is better for analyzing to join the capture instead of a screenshot.

Looking at the screenshot, I see that I didn't understand you first point correctly.
[...]

PR #549, opened 2021-03-02
Support Bootstrap-Read and Bootstrap-Discover

sbertin-telular:

### DIFF
--- a/core/bootstrap.c
+++ b/core/bootstrap.c
@@ -158,6 +158,356 @@ static void prv_requestBootstrap(lwm2m_context_t * context,
     }
 }
 
+static uint8_t prv_handleBootstrapDiscover(lwm2m_context_t * contextP,
+                                           lwm2m_uri_t * uriP,
+                                           uint8_t ** bufferP,
+                                           size_t * lengthP)
+{
+    size_t length;
+    size_t res;
+    int res2;
+    lwm2m_object_t * objectP;
+    uint8_t buffer[REG_OBJECT_MIN_LEN + 5];
+    lwm2m_list_t *instanceP;
+    lwm2m_uri_t uri;
+    int size;
+    lwm2m_data_t *dataP;
+    uint64_t val;
+
+    if (uriP && LWM2M_URI_IS_SET_OBJECT(uriP))
+    {
+        if (LWM2M_URI_IS_SET_INSTANCE(uriP)) return COAP_400_BAD_REQUEST;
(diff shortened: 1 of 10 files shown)


qleisan:
Starting to review (and read up on spec....). Expect to be done later today

qleisan (approved):
Code-wise I have no objections but I struggle a bit with the bigger picture (how this is used/triggered). I believe this code is already tested and "in production". (Right?) I'll approve so it can be merged but will continue to read up and do hands on tests

qleisan:
[210305_PR549_test1.pcapng.zip](https://github.com/eclipse/wakaama/files/6089303/210305_PR549_test1.pcapng.zip)
[210305_PR549_test2.pcapng.zip](https://github.com/eclipse/wakaama/files/6089305/210305_PR549_test2.pcapng.zip)

Found blocktransfer related issue while testing Bootstrap-Discover
modified `bootstrap_server.ini` as follows:
```
# For any client, we delete all server accounts and
# provision the Leshan server info
[Endpoint]
Delete=/0
Delete=/1
Server=1
Discover=/
```
launched example bootstrap server, server and client
```
./build-bootstrap/bootstrap_server -f examples/bootstrap_server/bootstrap_server.ini -4
./build-server/lwm2mserver -4
./build-client/lwm2mclient -4 -b -n bepa2
```

With `REST_MAX_CHUNK_SIZE` set to 128 (current setting) the result was according to **test2** - _ERROR_
With `REST_MAX_CHUNK_SIZE` set to 512 (modified setting) the result was according to **test1** - _OK_

also it would be helpful for the end user with some updated comments in the .ini file about the new command(s)

sbernard31:
About **210305_PR549_test2.pcapng.zip** this looks like #536, as Wakaama answers to a response with a response (See MID 55386).

I see that Bootstrap server try several time to send the discover request ? does it go in an infinite loop or the retry we see are just retransmission ?

sbertin-telular:
I'm not using these features in production. This was to make Wakaama more feature complete.

I'll update MAX_PACKET_SIZE  and add some comments to the .ini file.

qleisan (approved):

sbertin-telular:
Squashed

sbernard31:
@qleisan, @tuve about https://github.com/eclipse/wakaama/pull/549#issuecomment-791279205 we need maybe a dedicated issue to track this ?

qleisan:
created #551

Issue #550, opened 2021-03-08
observe_setParameters  may cause dump


lx121451:
if get a wrong msg，attrP->toSet be set to 0，we will get a dump in the last log api。

`LOG_ARG("Final toSet: %08X, minPeriod: %d, maxPeriod: %d, greaterThan: %f, lessThan: %f, step: %f",
            watcherP->parameters->toSet, watcherP->parameters->minPeriod, watcherP->parameters->maxPeriod, watcherP->parameters->greaterThan, watcherP->parameters->lessThan, watcherP->parameters->step);`
![wakaama](https://user-images.githubusercontent.com/20105247/110287188-0c165b80-8021-11eb-8336-570a751905a5.png)

sbertin-telular:
Thank you for reporting this. It does seem possible if the request is only clearing attributes with none set.

sbernard31:
Should I make it as a bug ?

sbertin-telular:
Yes.

qleisan:
Can confirm that a server clear command to a client built with debug and not having any attributes set to that URI causes a crash.
e.g. 
`clear 0 /31024/10/1`

sbernard31:
#552 integrated in master. This should be fixed now. 

@lx121451  do not hesitate to reopen or comment if you found anything wrong ;)

PR #552, opened 2021-03-08
Fix potential crash when manipulating object attributes

qleisan:
A server clean command on an URI with no attributes set caused client to crash (debug build)


qleisan:
fix for #550

sbertin-telular (approved):

PR #557, opened 2021-03-22
Fix debug step-timeout-print for 32-bit platforms

mlasch:
Current debug prints of timeout between lwm2m steps assume _time_t_ to be
64-bit. However, on MIPS and ARM this is not always given. This fix will decide
at build time whether a 32-bit or 64-bit print format is needed.

This was tested on:
 * Linux x86_64 (64-bit)
 * Linux armv5e (32-bit)
 * Linux mips32r2el-24kc-nf (32-bit)

Signed-off-by: Marc Lasch <marc.lasch@husqvarnagroup.com>

### DIFF
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -378,7 +378,7 @@ int lwm2m_step(lwm2m_context_t * contextP,
 {
     time_t tv_sec;
 
-    LOG_ARG("timeoutP: %" PRId64, *timeoutP);
+    LOG_ARG("timeoutP: %d", (int) *timeoutP);
     tv_sec = lwm2m_gettime();
     if (tv_sec < 0) return COAP_500_INTERNAL_SERVER_ERROR;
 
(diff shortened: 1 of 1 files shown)


sbertin-telular:
__TIMESIZE seems to be glibc specific. Since these are always timeouts in seconds, I would suggest LOG_ARG("timeoutP: %d", (int)*timeoutP);

core/liblwm2m.c
```diff
@@ -378,7 +378,8 @@ int lwm2m_step(lwm2m_context_t * contextP,
 {
     time_t tv_sec;
 
-    LOG_ARG("timeoutP: %" PRId64, *timeoutP);
+	LOG_ARG("timeoutP: %d", *timeoutP);
```
sbertin-telular, line 381: Needs a type cast to int. (int)*timeoutP

mlasch:
Changed it now. I was concerned there may be an issue on Windows, but I cannot test except with _compiler explorer_ (https://godbolt.org) where I get (with msvc 19.14):
`warning C4477: 'printf' : format string '%d' requires an argument of type 'int', but variadic argument 1 has type '__int64'`

sbertin-telular (approved):

Issue #555, opened 2021-03-19
Block transfer resend is ignored by client, halts transfer.


pehrnord:
https://github.com/eclipse/wakaama/blob/c5c6d321d50ff0534a390acc2bd9db7de48da41d/coap/block.c#L181

Client side: When receiving a resend of a previously handled block, it is ignored. This halts the block transfer, since the server will keep re-sending the block and then give up.
Triggered when client response to block transfer message is lost (?).
Should instead return COAP_231_CONTINUE?

sbernard31:
Is it possible to provide a wireshark capture to be sure that we understand the issue correctly.

qleisan:
[210323_555_recreation_attempt.pcapng.zip](https://github.com/eclipse/wakaama/files/6188131/210323_555_recreation_attempt.pcapng.zip)

> Client side: When receiving a resend of a previously handled block, it is ignored. This halts the block transfer, since the server will keep re-sending the block and then give up.
Triggered when client response to block transfer message is lost (?).

Attempted to recreate this issue without success, see attached wireshark log.
- client connects
- server writes data to client object, block transfer is triggered
- first client ACK (2.31) is dropped before it is processed by server (hack)
- server retransmits - everything works fine
- client disconnects

qleisan:
[210323_555_recreation_attempt_2.pcapng.zip](https://github.com/eclipse/wakaama/files/6189016/210323_555_recreation_attempt_2.pcapng.zip)

New attempt after code inspection, this time drop ACK from second (not first) block transfer packet. Problem recreated!

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #559, opened 2021-03-23
Fix blocktransfer stall due to lost ACK

qleisan:

Fix for #555

### DIFF
--- a/coap/block.c
+++ b/coap/block.c
@@ -247,8 +247,8 @@ uint8_t prv_coap_block_handler(lwm2m_block_data_t ** pBlockDataHead,
         }
 
         if (blockNum <= blockData->blockNum){
-            // this is a retransmissiion, ignore
-            return COAP_IGNORE;
+            // this is a retransmission
+            return COAP_RETRANSMISSION;
         }
 
         // If this is a retransmission, we already did that.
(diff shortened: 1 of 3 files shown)


coap/block.c
```diff
@@ -247,8 +247,8 @@ uint8_t prv_coap_block_handler(lwm2m_block_data_t ** pBlockDataHead,
         }
 
         if (blockNum <= blockData->blockNum){
-            // this is a retransmissiion, ignore
-            return COAP_IGNORE;
+            // this is a retransmission
+            return COAP_231_CONTINUE;
```
qleisan, line 251: what if this is the last block...?

qleisan:
Tested that the last ACK of the blocktransfer can be lost and will be resent as 2.04
Tested that an ACK in the middle of the blocktransfer can be lost and will be resent as 2.31

coap/block.c
qleisan: pushed new improved fix

rettichschnidi:
Any chance to get this fix unit-tested?

qleisan:
"Integration testing" is the only realistic way (as I see it) to get decent test coverage with realistic effort. And we should refactor some of the blocktransfer code once we have to make it cleaner and easier to understand.
I plan to look into integration testing now, happy to discuss best way forward

tuve (commented):
This looks fine to me. One question related to this, do we ACK retransmissions if they are block transfers (as we also should)?

sbernard31:
Shoul I integrate this one ?

qleisan:
yes and yes!

sbernard31:
Integrated in master (commit f4328b8f997ac7c8815f4f179695953cb193f5d3)

PR #563, opened 2021-03-26
Fix blocktransfer issue for bootstrap server

### LINKED ISSUE #551: bootstrap server answers a response with a response during block transfer
Issue described in #549, wireshark logs included

Direct links to important comments
https://github.com/eclipse/wakaama/pull/549#issuecomment-791279205
https://github.com/eclipse/wakaama/pull/549#issuecomment-791310453

qleisan:
Corrected handling of block2

This fixes #551, however it needs a thorough review!

### DIFF
--- a/examples/bootstrap_server/bootstrap_server.ini
+++ b/examples/bootstrap_server/bootstrap_server.ini
@@ -86,7 +86,8 @@ secret=4F7572536563726574
 [Endpoint]
 Delete=/0
 Delete=/1
-Server=2
+Server=1
+Discover=/
 
 # For the wakaama test client, we delete everything and
 # provision both the wakaama test server and the secure
--- a/include/liblwm2m.h
+++ b/include/liblwm2m.h
@@ -786,7 +786,8 @@ struct _lwm2m_context_
     void *                  monitorUserData;
 #endif
 #ifdef LWM2M_BOOTSTRAP_SERVER_MODE
-    lwm2m_client_t *        clientList;
+    lwm2m_client_t *        clientList;     // TODO: this is not used and should be removed??
+    lwm2m_block_data_t *       blockData;   // list to handle temporary block data.
     lwm2m_bootstrap_callback_t bootstrapCallback;
     void *                     bootstrapUserData;
 #endif
--- a/core/packet.c
+++ b/core/packet.c
@@ -727,11 +727,15 @@ void lwm2m_handle_packet(lwm2m_context_t * contextP,
                         peerP = utils_findBootstrapServer(contextP, fromSessionH);
                     }
 #endif
-#else
+#endif
+#ifdef LWM2M_SERVER_MODE
                     lwm2m_client_t * peerP;
                     peerP = utils_findClient(contextP, fromSessionH);
 #endif
-
+#ifdef LWM2M_BOOTSTRAP_SERVER_MODE
+                    lwm2m_context_t * peerP;
+                    peerP = contextP;
+#endif
                     if (peerP == NULL)
                     {
                         coap_error_code = COAP_500_INTERNAL_SERVER_ERROR;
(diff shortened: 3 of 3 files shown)


examples/bootstrap_server/bootstrap_server.ini
```diff
@@ -86,7 +86,8 @@ secret=4F7572536563726574
 [Endpoint]
 Delete=/0
 Delete=/1
-Server=2
+Server=1
```
qleisan, line 89: This edit is done to trigger block2 transfer to bootstrap server "out of the box". Could/should be removed in final version

> ./build-bootstrap/bootstrap_server -f examples/bootstrap_server/bootstrap_server.ini -4
> ./build-client/lwm2mclient -4 -b -n bepa2

examples/bootstrap_server/bootstrap_server.ini
```diff
@@ -86,7 +86,8 @@ secret=4F7572536563726574
...
+Discover=/
```
qleisan, line 90: This edit is done to trigger block2 transfer to bootstrap server "out of the box". Could/should be removed in final version

./build-bootstrap/bootstrap_server -f examples/bootstrap_server/bootstrap_server.ini -4
./build-client/lwm2mclient -4 -b -n bepa2

include/liblwm2m.h
```diff
@@ -786,7 +786,8 @@ struct _lwm2m_context_
     void *                  monitorUserData;
 #endif
 #ifdef LWM2M_BOOTSTRAP_SERVER_MODE
-    lwm2m_client_t *        clientList;
+    lwm2m_client_t *        clientList;     // TODO: this is not used and should be removed??
+    lwm2m_block_data_t *       blockData;   // list to handle temporary block data.
```
sbertin-telular, line 790: I think this might limit the server to only bootstrapping one client at a time. If the clientList isn't used, that might already be the case. Have you looked at that at all? If that is the case we should create an issue to correct it.

qleisan: I'm trying to wrap my head around it... I pushed this PR as a work in progress to get some feedback. 

Also having multiple identical "clientList" breaks the ability to build with both
`-DLWM2M_SERVER_MODE` and `-DLWM2M_BOOTSTRAP_SERVER_MODE` at the same, that I know some use.

qleisan: [210330_PR563_parallel_bootstrap.pcapng.zip](https://github.com/eclipse/wakaama/files/6229492/210330_PR563_parallel_bootstrap.pcapng.zip)
Practical testing show that bootstrapping works for two clients in parallel (attached wireshark capture)
```
./build-bootstrap/bootstrap_server -f examples/bootstrap_server/bootstrap_server.ini -4
./build-client/lwm2mclient -4 -b -n bepa2 -l 56830
./build-client/lwm2mclient -4 -b -n bepa3 -l 56831
```

sbertin-telular: I'm not convinced that the testing would show the problem. The first one sends the GET for the final block before the second one sends the first GET. The block requests aren't really interleaved. The data is also the same, so it might be hard to tell if there was a problem from what the server sees.

sbertin-telular: Eventually we should have blocktransfer support in the bootstrap server, but I'm fine waiting on that until other issues are resolved with the blocktransfer support.

tuve: I think we can hold off with this until we can do it right. One way might be to look into if we can refactor blocktransfer but that would be a bigger task.

mlasch: Getting block transfers right during bootstrap is a bigger task, since there is AFAIK no mechanism to track client contexts during bootstrap. So disabling block transfers during bootstrap seems sensible to me right now.

qleisan:
Closing this PR and will open an issue to add support for blocktransfer in bootstrap server

PR #566, opened 2021-03-30
Fix block2 response to a response bug

qleisan:

There is a general (not only bootstrap server related) issue with handling of block2 responses. Prefer to have this merged separately while working on a solution for the bootstrap server specific issue(s).

Without this fix a GET resulting in a block2 ACK will be answered with a 2.31 continue ACK...


mlasch (approved):
No faulty 2.31 continue ACKs seen anymore

sbernard31:
Integrated in master (commit 557c87b4fe5ac04d1d8ff25b4b1bde4ccb94386f)

Issue #182, opened 2016-06-24
Travsi-ci / any other CI


davidgraeff:
Hi,
some time ago I asked for this already (you remember, @dnav :D), but there was no definite answer. Is there any eclipse CI infrastructure that can be integrated into github, to execute the test suite on every commit and for every merge or can a public solution like travis-ci be used?

davidgraeff:
If the maintainers don't want a CI, this issue should be closed.

dnav:
It is more a lack of time than a lack of will. Sorry.

rettichschnidi:
I hacked together a simple PoC using GitHub Actions and SonarCloud: 
* [Repo overview](https://github.com/rettichschnidi/wakaama/tree/rettichschnidi-github-ci)
* [main.yml](https://github.com/rettichschnidi/wakaama/blob/rettichschnidi-github-ci/.github/workflows/main.yml)
* [Resulting report](https://sonarcloud.io/dashboard?id=rettichschnidi_wakaama) (Link will break in the future)

I have not configured SonarCloud, which would be very much needed to make it actually useful. Also, I very much like [Clang Static Analyzer](https://clang-analyzer.llvm.org/) and coverage-reports generated by [lcov](http://ltp.sourceforge.net/coverage/lcov.php). Before adding/rebasing to those tools, I'd like to know the following:

* Would a GitHub runner be acceptable to Eclipse?
* Who would enable SonarCloud access to this Wakaama repository? Maintain the SonarCloud account?

sbernard31:
>    Would a GitHub runner be acceptable to Eclipse?
>    Who would enable SonarCloud access to this Wakaama repository? Maintain the SonarCloud account?

You probably need to ask this to eclipse IT team opening a bugzilla issue : https://bugs.eclipse.org/bugs/enter_bug.cgi?product=Community

sbernard31:
About eclipse CI infrastructure, I think the classic eclipse way is jenkins : https://ci.eclipse.org/
(I don't if there is other way)

rettichschnidi:
I just created an issue in Bugzilla: https://bugs.eclipse.org/bugs/show_bug.cgi?id=568471

sbernard31:
Ok so I think we can use this bugzilla ticket to ask for a JIPP instance for Wakaama project.
[...]

PR #569, opened 2021-03-31
[GH-182] ci: Build with GCC and Clang

rettichschnidi:
This should give us some additional error detection.

Signed-off-by: Reto Schneider <code@reto-schneider.ch>


rettichschnidi:
@mlasch Pretty please review

mlasch (approved):
Tested with clang-10 and gcc-9

PR #570, opened 2021-04-01
[GH-182] ci: Run Clang Static Analyzer

rettichschnidi:
Findings are available as artifact, but no further actions (e.g. breaking the build) taken.

Currently, 79 bugs are found. Once those got resolved, I intend to prevent merging when the Clang static analyzer findings issues.

When reviewing this PR, please also check if this works properly on your local machine:
```
tools/ci/run_ci.sh --build --run-tests --scan-build scan-build-10
```
And review the results:
```
xdg-open build-wakaama/clang-static-analyzer/*/index.html
```

Please adapt `scan-build-10 ` in case you are not using the Clang 10 or install package `clang-tools-10` (on Debian).

PR #575, opened 2021-04-05
ci: Build using C99 and C11

rettichschnidi:
While the code base is still sprinkled with POSIX dependencies, this PR ensures the following:
* We do not use any GCC specific compiler extensions anymore
* Our code is compatible with C99

When reviewing, this should give you a couple of warnings (same as in master), but otherwise work:

```
tools/ci/run_ci.sh --clean --build --c-standard 11 --c-extensions ON
tools/ci/run_ci.sh --clean --build --c-standard 11 --c-extensions OFF
tools/ci/run_ci.sh --clean --build --c-standard 99 --c-extensions ON
tools/ci/run_ci.sh --clean --build --c-standard 99 --c-extensions OFF
```

Using this new version of run_ci.sh and the code from the current master, all builds with disabled extensions fail and the C99 ones throw additional warnings.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,41 @@
+name: C Standards
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      matrix:
+        compiler: ["clang-10", "gcc-10"]
+        c_standard: ["99", "11"]
+        c_extensions: ["ON", "OFF"]
+
+    steps:
+    - name: Checkout code including full history and submodules
+      uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


rettichschnidi:
@qleisan Added some description about how to test the PR.

qleisan (approved):

sbernard31:
> I will rebase on master once #570 got merged.

It's integrated.  :-)

sbernard31:
Should I wait action was no more stuck  on queued state before to integrate this in master ?

rettichschnidi:
> Should I wait action was no more stuck on queued state before to integrate this in master ?

I think so, yes. We should not merge stuff that has not passed all tests.

Issue #494, opened 2020-11-06
Supported compilers, architectures, language versions


rettichschnidi:
I could not find any documentation regarding targeted (supported) compilers/architectures/languages/etc. Did I miss something?

If not, my needs/proposal would be:
* Compilers:
  * Latest version of GCC
  * Latest version of Clang
* Architectures:
  * x86_64 on Linux
  * MIPS 32 bit on Linux
  * ARM 32 bit on Linux
* Language versions:
  * C99/C11: Fully supported
  * C90: Only for what might become a 1.x maintenance branch (#487). No longer supported in master, relieving the project from its restrictions.
  * C++: While Wakaama contains no C++ code, it must be usable from a C++ project.

sbertin-telular:
My primary use is on a non-Linux system using the IAR compiler. It isn't practical to add my environment as supported, but I would not like to see any gcc/clang extensions used or Linux assumed outside of the examples. I would like for this to continue to be usable on constrained devices where running Linux isn't possible or practical.

tuve:
My primary use is also on a non Linux system. On the embedded devices i think we run it with free rtos or bara bone on the mcu and then we also run it as a server as a library on iOS. I think there is also some (non linux) desktop use for testing as well.

rettichschnidi:
> My primary use is on a non-Linux system using the IAR compiler. 

Which architecture? ARM? Something with less than 32 bits?

> It isn't practical to add my environment as supported, but I would not like to see any gcc/clang extensions used or Linux assumed outside of the examples.

Regarding the extensions: How about adding a build configuration to CI which builds the "library" parts using -Wpedantic and "-Werror"?
Regarding non-Linux targets: Do you need Wakaama, client and/or server, to work with freestanding C implementations? Or is there some kind of OS?

sbertin-telular:
I'm currently using an Arm Cortex-M3 based system, so 32-bit.

A CI build of the library parts with -Wpedantic and -Werror would probably be helpful.

I'm currently using an RTOS, but would like to see it continue to be possible to use it without an OS. As an example, adding mutexes (provided by the implementer) for thread safety would be OK, but requiring multiple threads would not. The mutex functions could then do nothing in a no-OS environment.

rettichschnidi:
PR #575 will ensure Wakaama can be built with C99 (and C11) as well as keeping GNU specific compiler extensions out of the code base.

tuve:
@rettichschnidi @sbernard31 can we close this?

[...]

Issue #544, opened 2021-02-22
Wakaama needs a more abstract API


mlasch:
In order to simplify the usage and lower the entry barrier to Wakaama, I think the way of how to interact with library data structures should be reworked.
E.g. currently, on the server side, each implementation (examples and custom builds) have to handle the raw buffer. The buffer can contain all different content formats and parsing has to be implemented for each new application.
There is already the `lwm2m_data_t` structure in place which can hold IPSO objects independently of the content format in a tree structure. Also functions like `lwm2m_data_parse` and `lwm2m_data_serialize` are available. However they are a bit cumbersome to use, partly because in addition to the pointer to the structure there is also the size variable which have to be tracked. A clean interface would help a lot.

Some ideas:
* Modify the result callback to hide the raw buffer and do the parsing before the call (this assumes that all chunks of a block transfer can fit in memory):
```
typedef void (*lwm2m_result_callback_t) (uint16_t clientID, lwm2m_uri_t * uriP, int status, block_info_t * block_info, lwm2m_media_type_t format, uint8_t * data, int dataLength, void * userData);
```
<->
```
typedef void (*lwm2m_result_callback_t) (uint16_t clientID, lwm2m_uri_t * uriP, int status, block_info_t * block_info, lwm2m_data_t * lwm2m_data, int size, void * userData);
```
* Use a single structure which holds all important data like uri, parsed tree, size, status etc.
* As described in #521 the interface is already misused to report the notify counter via the status field.
* The callback should still allow access to the raw buffer if needed
  * If the chunks of a block transfer have to be treated individually
  * If parsing fails and custom parsing is needed
* It should be possible to disable the parsing (at build time?) due to memory concerns

This may imply major changes, so this issue is intended to collect and discuss ideas on how to improve the overall usability. Any thoughts?

sbernard31:
@sbertin-telular, @qleisan, @tuve do you think some of this idea makes sense ? 

@mlasch if dev team agree with some of your ideas do you plan to contribute it or you just share the idea ?

qleisan:
Yes definitely. The discussion should also cover how to interact more easily with objects on the client side from the "application".

sbertin-telular:
On the client side one thing that would be helpful is to allow writes to be local or remote. Local writes might have more access. For example, assume a resource that holds the current temperature. This should be writable locally, but not remotely. Right now a separate function is needed to modify the value in the object when a new reading is made.

[...]

PR #577, opened 2021-04-05
[GH-494] ci: Build and test on multiple GNU/Linux architectures

rettichschnidi:
Foundation for coverage of non-amd64 GNU/Linux architectures in our CI. Not using qemu-user directly on Ubuntu 20.04 as it is missing support for certain architectures which are supported by Debian (armv6/armel, mipsel, and others).

Some unit tests where failing on aarch64, s390x and ppc64le before. Fixed it by rewriting utils_textToFloat(). See commit messages for details.

Please note:
* Due to [incompatibilities of CMake 3.16, which is bundled with Debian Buster, and QEMU](https://gitlab.kitware.com/cmake/cmake/-/issues/20568), which executes the Docker containers with the foreign architectures, updating to CMake version 3.19 or newer is needed. As KitWare only offers prebuilt binaries for x86_64 and aarch64, we need to build CMake from source, which takes a long time (~90 minutes). Luckily, the resulting docker image will be cached and subsequent runs (on the same GitHub account) run much faster.

At a later point, I intend to also add GNU/Linux on MIPSel and, if possible, ARMv5. Before this can happen however, those architectures must be added to https://github.com/uraimo/run-on-arch-action first.

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -136,114 +136,37 @@ int utils_textToFloat(const uint8_t * buffer,
                       double * dataP,
                       bool allowExponential)
 {
-    double result;
-    int sign;
-    int i;
+    int ret = 0;
 
-    if (0 == length) return 0;
-
-    if (buffer[0] == '-')
-    {
-        sign = -1;
-        i = 1;
-    }
-    else
-    {
-        sign = 1;
-        i = 0;
+    if (length == 0) {
+        return 0;
     }
 
-    /* Must have a decimal digit first after optional sign */
-    if (i >= length || !isdigit(buffer[i])) return 0;
+    char *const buffer_c_str = lwm2m_malloc(length + 1);
+    char *tailptr;
 
-    result = 0;
-    while (i < length && buffer[i] != '.' && buffer[i] != 'e' && buffer[i] != 'E')
-    {
-        if (isdigit(buffer[i]))
-        {
-            if (result > (DBL_MAX / 10)) return 0;
-            result *= 10;
-            result += (buffer[i] - '0');
-        }
-        else
-        {
-            return 0;
-        }
-        i++;
+    if (!buffer_c_str) {
+        return 0;
     }
-    if (i < length && buffer[i] == '.')
-    {
-        double dec;
 
-        i++;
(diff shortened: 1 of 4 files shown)


rettichschnidi:
Rebased on most recent master, knowing this will break the build on aarch64, s390x and ppc64le. Its just too good an example of why this PR makes sense. :)

Will submit a fix later on.

rettichschnidi:
> Rebased on most recent master, knowing this will break the build on aarch64, s390x and ppc64le. Its just too good an example of why this PR makes sense. :)
> 
> Will submit a fix later on.

Done. No longer a draft -> open for reviews :smile:

tuve (approved):
Looks fine to me, but i find it a bit odd that the function is named utils_textToFloat and then we do strtod and populate a a double. Maybe it is nitpicking but shouldn't yo rename the function also (this will of course affect everywhere it is used).

rettichschnidi:
> Looks fine to me, but i find it a bit odd that the function is named utils_textToFloat and then we do strtod and populate a a double. Maybe it is nitpicking but shouldn't yo rename the function also (this will of course affect everywhere it is used).

I think the current API needs a lot of rework, including an upfront discussion (#544?), but I'd rather not do this as part of this issue.

core/utils.c
```diff
@@ -136,114 +136,37 @@ int utils_textToFloat(const uint8_t * buffer,
...
-    {
-        sign = 1;
-        i = 0;
+    if (length == 0) {
+        return 0;
     }
 
-    /* Must have a decimal digit first after optional sign */
-    if (i >= length || !isdigit(buffer[i])) return 0;
+    char *const buffer_c_str = lwm2m_malloc(length + 1);
```
sbertin-telular, line 145: I realize this has already been merged, but I would prefer to see utils_textToFloat operate without the dynamic memory allocation. Dynamic memory adds a possible failure that is very real on a constrained device.

rettichschnidi: Would have preferred it too, but could not see a realistic way to get the conversion working properly (i.e. passing the existing unit tests) on all architectures.

PR #576, opened 2021-04-05
ci: Break build on unit test error

rettichschnidi:
With this commit, failing tests will break the build.

To test this locally, modify the code so that at least one test fails, then run this:
```bash
tools/ci/run_ci.sh --all
```

You should see the failing tests:
```
Run Summary:    Type  Total    Ran Passed Failed Inactive
              suites      6      6    n/a      0        0
               tests     65     65     64      1        0
             asserts    711    711    710      1      n/a

Elapsed time =    0.000 seconds

  1. /home/reto/code/3rd-party/wakaama/tests/block1tests.c:45  - CU_ASSERT_EQUAL(bsize,1)
```

The exit code should be 1:
```bash
$ echo $?
1
```

PR #579, opened 2021-04-06
Update/cleanup cmake files

mlasch:
This is a pre-work/cleanup to make the CoAP block-size configurable during build time (via cmake variable) in a future PR.
 - Replace redundant WITH_LOG since it was only rarely used and we have LWM2M_WITH_LOG.
 - Replace add_definitions with modern alternatives.
 - Set a common minimal required cmake version.

I bumped the cmake version in all files to 3.13 to support all required features and to be consistent (3.13 is currently available in Debian stable, so still quite conservative). Also tried to make the formatting a bit more homogeneous.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.13)
 
 add_compile_definitions(_POSIX_C_SOURCE=200809)
 add_compile_options(-pedantic)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,6 +1,6 @@
-cmake_minimum_required (VERSION 3.13)
+cmake_minimum_required(VERSION 3.13)
 
-project (lwm2munittests C)
+project(lwm2munittests C)
 
 add_compile_definitions(_POSIX_C_SOURCE=200809)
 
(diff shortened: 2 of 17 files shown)


tests/CMakeLists.txt
```diff
@@ -7,15 +7,18 @@ include(${CMAKE_CURRENT_LIST_DIR}/../coap/coap.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/../data/data.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/../examples/shared/shared.cmake)
 
-add_definitions(-DLWM2M_CLIENT_MODE -DLWM2M_SUPPORT_TLV -DLWM2M_SUPPORT_JSON)
+add_compile_definitions(LWM2M_CLIENT_MODE)
```
rettichschnidi, line 10: I like the idea of using multiple lines.

rettichschnidi (approved):

rettichschnidi:
FYI: This will conflict with PR #576. One of us will have to rebase.

examples/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.13)
```
rettichschnidi, line 1:
```suggestion
cmake_minimum_required (VERSION 3.13)
```

qleisan: Why the extra space?

rettichschnidi: For consistency reasons - it has this space everywhere else too. But really not important, which is why have approved the PR anyway.

sbernard31:
Could you rebase on master ? :pray:

PR #581, opened 2021-04-07
doc: Add badges

rettichschnidi:
Initial set of badges. To be extended once we have services integrated which provide more shiny badges.

Preview can be seen here: https://github.com/husqvarnagroup/wakaama/tree/gardena/rs/add-badges


qleisan (approved):
LGTM

PR #582, opened 2021-04-10
Add missing context to raw block transfer function calls.

pehrnord:
Signed-off-by: Pehr Nord <pehr.nord@husqvarnagroup.com>

### DIFF
--- a/core/objects.c
+++ b/core/objects.c
@@ -538,7 +538,7 @@ uint8_t object_raw_block1_write(lwm2m_context_t * contextP,
     // TODO: support resource instance
     if (LWM2M_URI_IS_SET_RESOURCE_INSTANCE(uriP)) return COAP_400_BAD_REQUEST;
 #endif
-    result = targetP->rawBlock1WriteFunc(uriP, format, buffer, length, targetP, block_num, block_more);
+    result = targetP->rawBlock1WriteFunc(contextP, uriP, format, buffer, length, targetP, block_num, block_more);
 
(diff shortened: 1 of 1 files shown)


rettichschnidi (changes requested):
Thanks for the input!

My feedback:
* The error being fixed here could only get into the code base because LWM2M_RAW_BLOCK1_REQUESTS was never enabled in any of the targets. In order to merge this PR, this should be addresses.
<s>I know @mlasch is working on something in this area - maybe he wants to chime in?</s>
* You need to sign the commits using an e-mail address which is signed up at Eclipse and has accepted the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php).

pehrnord:
Signed up and accepted ECA.
Not sure how to proceed on you other comment, would a new unittest need to be added?

rettichschnidi:
> Signed up and accepted ECA.

Seems to be working now. :smile: 

> Not sure how to proceed on you other comment, would a new unittest need to be added?

Having unit tests for the code in question would be very nice, but seems a bit too much to ask for in the context of this PR.

To prevent this specifc issue from happening again we'd need to make sure at least one target gets built with this define (LWM2M_RAW_BLOCK1_REQUESTS) enabled. And also, that at least one keeps getting built without this flag.

I think the best way to ensure this would be to create a new build job which compiles the sources both ways and lets the unit tests run. I opened up a new issue #588 for this, and I plan on implementing it soon.

Therefore, I will merge this PR.

Thanks a lot and welcome to the project!

rettichschnidi (approved):

PR #602, opened 2021-05-06
Add command line option to set default format

qleisan:
New client switch '-f' sets default content format


sbertin-telular:
Is there a use case for this other than testing? I would expect this to be known at compile time for most constrained devices and I'm concerned about growing the code size unnecessarily.

qleisan:
No this is for testing and I pushed it as a draft to get feedback. Didn't turn out as elegant as I hoped. Integration testing (see other draft - your feedback is much appreciated) would benefit greatly if the default format could be set as an argument. Any suggestions that would improve the situation? 
The integration tests aim to follow this document [Enabler Test Specification (Interoperability) for Lightweight
M2M](https://www.openmobilealliance.org/release/LightweightM2M/ETS/OMA-ETS-LightweightM2M-V1_1-20190912-D.pdf)

qleisan:
@rettichschnidi code formatting (clang-10) rules force new/modified code to be updated but the rest of the code in the file might be following a different style. Usually one should keep to the style in the file you make changes. Any thoughts on this? Do another commit that reformat the entire file according to our current formatting rules?

core/registration.c
```diff
@@ -1330,7 +1330,6 @@ static uint16_t prv_splitLinkAttribute(uint8_t * data,
 
     index--;
     while (index > *valueStart && data[index] == ' ') index--;
-    if (index == *valueStart) return 0;
```
sbertin-telular, line 1333: It isn't clear to me why this was removed. It would allow a missing value to be treated as a '\0', REG_ATTR_SEPARATOR, or ' ' depending on what was in the buffer,

sbertin-telular:
> No this is for testing

I just had a general concern about unnecessarily growing the code size in core/*. This doesn't seem too bad, but if we add several things that aren't too bad we end up with code bloat.
> Any suggestions that would improve the situation?

The only alternative I see would be different builds with different formats. That would also more closely match actual use cases. Not sure how practical that would be to implement and test.

core/registration.c
qleisan: Will have to revisit this, the issue was that "-f 0" was not accepted (passing a zero as the default format)

rettichschnidi:
> @rettichschnidi code formatting (clang-10) rules force new/modified code to be updated but the rest of the code in the file might be following a different style. Usually one should keep to the style in the file you make changes. Any thoughts on this? Do another commit that reformat the entire file according to our current formatting rules?

The intended way to deal with this would be to create a pure-cosmetic patch which reformats the needed sections with clang-format (clang-format can format files just partially). Once done, implement your changes, which then will no longer trigger reformatting any existing code.

This way we can work towards a sanely formatted code base step-by-step.

qleisan:
@rettichschnidi thanks for the review comments. We must address the bigger issue highlighted by @sbertin-telular also

> Is there a use case for this other than testing? I would expect this to be known at compile time for most constrained devices and I'm concerned about growing the code size unnecessarily.

Is adding another command line switch (one that is only for test purposes) in the client not acceptable? Is there a more elegant way to do it than I have suggested in this draft? I'm not to happy about the implementation approach myself since it breaks a pattern. Constructive feedback welcome.

core/objects.c
```diff
@@ -798,7 +798,11 @@ int object_getRegisterPayloadBufferLength(lwm2m_context_t * contextP)
         index += strlen(REG_DEFAULT_PATH);
     }
 
-    index += strlen(REG_LWM2M_RESOURCE_TYPE);
+    if (contextP->defaultregformat != NULL) {
+        index += strlen(">;rt=\"oma.lwm2m\",") + 4 + strlen(contextP->defaultregformat);
```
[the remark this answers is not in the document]
tuve: or a constant

sbernard31:
I implemented support of "ct=" attribute in Leshan recently. (https://github.com/eclipse/leshan/pull/1006 and https://github.com/eclipse/leshan/pull/1008)

And my understanding of this feature is that ct attribute does not contain the "default format" but rather "all optional supported formats" by the device.

qleisan:
I'm looking into an alternative approach - building custom builds of the example client for integrationtests that require json/tlv/etc.

PR #598, opened 2021-04-28
Add integration tests

qleisan:
Add initial pytest based integration tests

PR #603, opened 2021-05-07
Add integration test int-104 (requires PR#598 and PR#561)

qleisan:
Test that the Client updates its registration with the Server when triggered with the Registration Update Trigger

### DIFF
--- a/tests/integration/test_registration_interface.py
+++ b/tests/integration/test_registration_interface.py
@@ -1,0 +1,62 @@
+import re
+import pytest
+
+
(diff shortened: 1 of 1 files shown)


rettichschnidi:
This is hard to review - would rather see this merged into #598.

rettichschnidi:
I do not think this is the right approach - I'd rather have #598 massaged really well - just one (or very few tests), but including CI integration, before adding many PRs which depend on #598.

qleisan:
Reorganizing - will create a new PR with this and other content

PR #604, opened 2021-05-07
Add integration tests int-201, int-204 (requires PR#598, PR#602)

qleisan:
Add integration tests that use plain-text and JSON

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,118 @@
+import pytest
+import pexpect
+
+
+class HelperBase:
+    def __dumptext(self):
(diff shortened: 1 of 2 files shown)


rettichschnidi:
This is hard to review - would rather see this merged into #598.

qleisan:
I just split #598 to reduce the number of dependencies to get this first integration test PR merged to master

rettichschnidi:
I do not thing this is the right approach - I'd rather have #598 massaged really well - just one (or very few tests), but including CI integration, before adding many PRs which depend on #598.

qleisan:
Reorganizing - will create a new PR with this and other content

Issue #607, opened 2021-05-20
Example client doesn't include lifetime in registration update


qleisan:
Integration test **int-107** [Enabler Test Specification (Interoperability) for Lightweight M2M](https://www.openmobilealliance.org/release/LightweightM2M/ETS/OMA-ETS-LightweightM2M-V1_1-20190912-D.pdf) specifies that _Client immediately Updated its Registration with the payload indicating that Lifetime has changed, i.e. CoAP POST on /rd/{location}?lt=120_ after a write (CoAP PUT) on /1/x/1. Current implementation doesn't supply the `?lt=120` part

qleisan:
Draft implementation of int-107 integration test case (removed from PR#611)

```
@pytest.mark.slow
@pytest.mark.client_args("-t 60")
def test_extending_lifetime_registration(lwm2mserver, lwm2mclient):
    """LightweightM2M-1.1-int-107
    Test that Server can extend the LwM2M lifetime of Client

    Test only partially implemented due to limitations (see comments)"""

    # Test Procedure 1
    # Pass-Criteria A
    lwm2mclient.waitfortext("STATE_READY")
    # Test Procedure 2
    assert lwm2mserver.commandresponse("write 0 /1/0/1 120", "OK")
    text = lwm2mserver.waitforpacket()
    # Pass-Criteria B
    # a)
    assert text.find("COAP_204_CHANGED") > 0
    # b) NOT IMPLEMENTED
    # c) not observable
    # Test Procedure 3
    ...
```

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #611, opened 2021-05-25
Add more integration tests

qleisan:
Add int-102, int-105, int-107, int-301, int-302


mlasch (changes requested):

tests/integration/test_information_reporting_interface.py
```diff
@@ -0,0 +1,60 @@
...
+    assert lwm2mserver.commandresponse("time 0 /3/0/7 5 15", "OK")
+    assert lwm2mserver.commandresponse("time 0 /3/0/8 10 20", "OK")
+    # Test Procedure 3
+    assert lwm2mserver.commandresponse("observe 0 /3/0/7", "OK")
+    assert lwm2mserver.commandresponse("observe 0 /3/0/8", "OK") #consumes first observe from /3/0/7
+    text = lwm2mserver.waitforpacket() # consumes ACK and piggy-backed first observe from /3/0/8
+    text = lwm2mserver.waitfortime(20) # wait for max "pmax"
+    # Pass-Criteria A
+    assert text.count("Notify from client #0 /3/0/7") > 0
+    assert text.count("Notify from client #0 /3/0/8") > 0
```
mlasch, line 31: Would it be feasible to check here also the values of Power Source Voltage (7) and Power Source Current (8)? Would be nice to check if the notification messages actually contain the observed values.

tests/integration/test_registration_interface.py
```diff
@@ -42,3 +43,87 @@ def test_registration_interface(lwm2mserver, lwm2mclient):
...
+    assert text.find("COAP_204_CHANGED") > 0
+    # b) NOT IMPLEMENTED
+    # c) not observable
+    # Test Procedure 3
+    text = lwm2mserver.waitfortime(120)
+    # Pass-Criteria C
+    # one update is sent immediately, expect implementation
+    # to send update at lifetime/2 intervals
+    # this does not seem to be the case for 120s (more frequent)
+    assert text.count("Client #0 updated.") >= 2
```
mlasch, line 129: Why not test for exactly two updates during the lifetime? This however interferes with #613 when testing against 120s.

tests/integration/conftest.py
```diff
@@ -46,6 +48,13 @@ def waitfortext(self,txt):
             return False
         return True
 
+    def waitfortime(self, timedelay):
+        '''Wait for time and return output since last command'''
```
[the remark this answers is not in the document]
qleisan: will fix

tests/integration/test_information_reporting_interface.py
qleisan: Will add payload check

tests/integration/test_registration_interface.py
```diff
@@ -42,3 +43,87 @@ def test_registration_interface(lwm2mserver, lwm2mclient):
...
+    # wait long enough to client to send register update (180s total)
+    # Pass-Criteria C
+    # only in debug build it is possible to detect "4.04" from printouts
+    # Pass-Criteria D
+    # Client should perform full registration according as in int-101
+    # this is not implemented, client exits with "STATE_BOOTSTRAP_REQUIRED"
+
+
+@pytest.mark.slow
+#@pytest.mark.client_args("-t 60")
```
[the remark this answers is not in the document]
qleisan: The entire testcase needs to be removed due to #607

tests/integration/test_registration_interface.py
qleisan: TC removed. (The exact number of updates is implementation dependent)

mlasch (approved):

PR #627, opened 2021-09-27
off-by-one: block transfer fix

broglep-work:
In case of `payload_len == x * block_size` wakaama did send an extra / erroneous block1 with `length == 0` after it already did send the actual last block with CoAP header more=0


tuve (approved):

tuve:
Just waiting for all the checks to pass, then I think we can merge this.

mlasch (approved):

PR #628, opened 2021-09-27
Fix payload buffer use after free

broglep-work:
The caller of a method (e.g. `lwm2m_dm_write`) could have free'd the buffer after
the method completes (assuming that he retains ownership of allocated memory).
But the payload buffer is used by wakaama even beyond the scope of the method /
current transaction. This happens for block transfers but could also happen in
case of re-transmissions. For such cases, it is necessary for wakaama to copy
the payload for the transaction.


mlasch (approved):
Looks good to me.

tuve (approved):
Looks good to me

PR #630, opened 2021-09-28
fix: errors in transaction handling

broglep-work:
- ` prv_get_transaction` did return non-matching transaction if only one
  transaction was in the list
- `prv_checkFinished` did not check if the `mid` did match, resulting in a wrong
  transaction to be removed in `transaction_handleResponse`


broglep-work:
Seems to have some other undesirable side effects, need to have a look at those first

PR #631, opened 2021-09-30
fix: errors in transaction handling

broglep-work:
- `prv_get_transaction` did return non-matching transaction if only one
  transaction was in the list or only session or mid did match
- `prv_checkFinished` did not check if the mid did match, resulting in a wrong
  transaction to be removed in `transaction_handleResponse`


mlasch (approved):

mlasch:
This is an important fix, @tuve do you also want to have a look?

tuve (approved):

Issue #618, opened 2021-06-09
coap_packet_t - type of payload_len member


piotr276:
Hello,

Using wakaama client from examples I am implementing some basic firmware update functionality. I have added my implemenation to the "write" method of the firmware object. Sending files from the leshan server to my client is working fine, but surprisingly the maximum file size is limited to ~ca. 65 kB (u16 payload size). Is this intended?

I can see that in the **coap_packet_t** struct, the member payload_len is defined as uint16_t. But in many other places in wakaama client example code, the payload_len is defined as int or size_t.

For example:

`void lwm2m_handle_packet(lwm2m_context_t* contextP, uint8_t* buffer, int length, void* fromSessionH);`

is using "length" as int and in this function we have casting from int to u16 which looks strange to me:

packet.c -> line 476

`coap_error_code = coap_parse_message(message, buffer, (uint16_t)length);`

Could you be so kind and provide some clarification why the payload_size is defined in **coap_packet_t**  as u16? Is this a requirement of CoAP protocol?

After I have changed this u16 to u32, everything working fine (at least in my "playground") and I am able to send files larger than ~65 kB succesfully.

piotr276:
You are right, but in the wakaama client -> the CoAP block size is set by defautl to 1024, so anyway for files larger than 1024 the block-wise transfer is triggered. IMO this should be possible also for files larger than 65 kB, but it is not due to this limitation in coap_packet_t (or I am doing something wrong - I am using "write" from the leshan demo server so the "write" method from client's firmware object is called - in my simple implementation I just copy the buffer content to the file). Without any changes in coap_packet_t struct for files larger than 65 kB there is an COAP_406_NOT_ACCEPTABLE in response for write operation from leshan server which is because there is a casting from int to u16, so the payload size is truncated and due to that, wrong size is passed further which ends in fail in one of if statements.

mlasch:
@piotr276 you're doing it right, this is also an issue on the server side. Writes larger than 65kB will be truncated because of the implicit type conversion. No error is reported, the write will just finish ordinarily.

My suggestion would also be to increase the payload_len to `uint32_t` which would allow larger writes. The UDP packet size limit should not be an issue because block transfer will be used anyways as you said above.

tuve:
I agree we should change the payload_len to size_t, this is just an oversight that it did not change when we introduced block transfer.

mlasch:
@piotr276  I close this issue, as the fix was merged.

PR #632, opened 2021-10-04
[#618] Fix payload_len handling

mlasch:
Use size_t/ssize_t to store the message buffer sizes instead of a uint16_t and int types.

This fix allows transactions larger than 65k as discussed in #618.


tuve (approved):
Looks good to me

PR #626, opened 2021-09-25
Integration of Mbed TLS for lwm2mclient example

LukasKarel:
As mentioned in Issue #623 I can provide a sample implementation with mbedtls. Please remark this implementation was done quick and dirty with some files I had around. It was not implemented to be merged on master! I used the networklayer (mbedtls_net_context) shipped with mbedtls. Currently only PSK mode is supported. I tested it with the leshan server.

This implementation is not secure and I am sure it does not compile with tinydtls anymore.

I hope it works without greater problems. Otherwise feel free to ask me anything about it.

PR #646, opened 2022-01-23
CI: Improve compilation coverage

rettichschnidi:
This builds the tinydtls-enabled example client in CI.

This is the first step in making it easy for #626 to implement some kind of testing.


rettichschnidi:
Got some new ideas, will rework this.

Issue #487, opened 2020-11-03
Future of 1.0 version


sbernard31:
Wakaama [`1.0`](https://github.com/eclipse/wakaama/tree/v1.0) is an implementation of LWM2M 1.0.x

Current `master` branch aims to target a LWM2M 1.1.x.
This`master` branch will probably break `1.0` API and so will be released as a 2.0 version (or higher if needed)

The question is :  **What should we do with the `1.0`, do we plan to still maintain it (at least bug fix release) ?**

@sbertin-telular, @rettichschnidi any though about this ?

Feedback from community would also be really appreciated :pray:

rettichschnidi:
Personally, I have no interest in maintaining a 1.x release. Catching up with LWM2M 1.1+ and not being constrained by the existing API seems much more attractive to me.

sbertin-telular:
Unless there is strong community demand I also don't have interest in maintaining the 1.0 branch.

jvermillard:
On sierra side, I don't mind breaking the API, but having the client being able to support both 1.0 and 1.1 (if feasable at all) would be interesting since I suspect we will need to handle both cases in the wild

sbertin-telular:
My 1.1 changes have #defines to select which LWM2M version will be supported. This could be important for more constrained devices to limit memory requirements if only 1.0 features are required.

tuve:
I think the 1.0 version needs to be maintained, at least for bug fixes. On
really constrained devices it might not be possible to migrate to LWM2M
1.1.x with a (likely) increased memory footprint just to get a bugfix. Also
The LWM2M 1.0.x support is not complete in the Wakaama 1.0 release, object
versioning support as an example and things like the block transfer support
is an addition to coap and does not depend on the lwm2m version (the later
should probably go into version 2.0)

Den tis 3 nov. 2020 kl 16:45 skrev Simon <notifications@github.com>:

sbernard31:
I added Security policy (https://github.com/eclipse/wakaama/issues/491#issuecomment-722258006)
This thread is probably the good place to deciding which version is concerned by this policy.
The idea is not have something like : 
| Version | Supported          |
| ------- | ------------------ |
| 5.1.x   | :white_check_mark: |
| 5.0.x   | :x:                |
| 4.0.x   | :white_check_mark: |
| < 4.0   | :x:

jvermillard:
> Maybe do as @sbertin-telular with defines is enough?
[...]

PR #650, opened 2022-01-30
[GH-487] Declare past releases as unsupported

rettichschnidi:
Nobody stepped up to assemble a new 1.0 based maintenance/security
release. At the very least our users should know that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,17 @@ protocol (LWM2M).
 
 Developers mailing list: https://dev.eclipse.org/mailman/listinfo/wakaama-dev
 
+## Security warning
+
+The only official release of Wakaama, version 1.0, is affected by various
+security issues ([CVE-2019-9004], [CVE-2021-41040]).
+
+Please use the most recent commit in the master branch. Release 1.0 is not
+supported anymore.
+
+[CVE-2019-9004]: https://www.cve.org/CVERecord?id=CVE-2019-9004
+[CVE-2021-41040]: https://www.cve.org/CVERecord?id=CVE-2021-41040
+
 ## Badges
 [![Build](https://github.com/eclipse/wakaama/actions/workflows/build_and_test.yaml/badge.svg)](https://github.com/eclipse/wakaama/actions/workflows/build_and_test.yaml)
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,4 +10,7 @@ For more details, please look at https://www.eclipse.org/security/.
 
 ## Supported Versions
 
-To Be Defined.
+None
+
+Please use the most recent commit in the master branch. Release 1.0 is not
+supported anymore.


mlasch (approved):

rettichschnidi:
@qleisan @tuve @sbernard31 any words from you? Would rather not merge something like this without getting your inputs.

tuve (approved):

tuve:
@rettichschnidi I think you should merge this and close #487 since it is the de facto state.

sbernard31:
As I said at (https://bugs.eclipse.org/bugs/show_bug.cgi?id=577968#c13) : 

> Generally, I would say that at least the last stable release should get bug fixes / security fixes. But Wakaama project situation is a bit different as the project was abandoned and is in a revival process. So I could understand that the team prefer to focus on making the project back on track and so don't invest much time in very old release.
>
>  You should also consider your perception of current community needs to take this decision. Does the community ask for this kind of 1.x support or does everybody seems to switch to `master`.
> 
> Note that I feel it's OK to say you don't support 1.x version, then change your mind later if there is strong community request for this and **if the team is able to support this.**

From https://github.com/eclipse/wakaama/issues/487#issuecomment-1019356446, nobody seems to react , so I understand your choice :slightly_smiling_face: 
Note this is only 16 days ago when I write this line, so maybe you should wait a little bit ? (expect if you already know that you will really not able to handle this support)

rettichschnidi:
> Note this is only 16 days ago when I write this line, so maybe you should wait a little bit ? (expect if you already know that you will really not able to handle this support)

While backporting those specific CVE fixes to 1.0 is certainly, I can not justify the time needed to do crank out a proper release (tested, all know issues fixed, etc). I'd rather have no release than one that gives the user a false of security.

What bothers me a bit [is this](https://github.com/eclipse/wakaama):

image

I have not read up if and how lightweight a snapshot release could be done within the Eclipse rule set, but maybe we could create one to make the 1.0 release less prominent?

sbernard31:
> I have not read up if and how lightweight a snapshot release could be done within the Eclipse rule set, but such one might make the 1.0 release less prominent.

For milestones release, AFIAK there is nothing special. 
At least for Leshan we just to a release without any eclipse process. 
Indeed, this will allow you to hide the 1.0.0 version.
 
In Leshan we have this kind of [documentation](https://github.com/eclipse/leshan/blob/master/SECURITY.md) and I add this kind of banner : 
### :exclamation::skull_and_crossbones::exclamation:   [**Affected by security issue**](https://github.com/eclipse/leshan/security/policy#versions-security-state):exclamation::skull_and_crossbones::exclamation: 
To all affected milestone or stable release, e.g :  https://github.com/eclipse/leshan/releases/tag/leshan-2.0.0-M3

sbernard31:
Eventually you can also add a big warning about the 1.0.0 version in the README with a link to the reason behind this and what user should do ?

rettichschnidi:
Updated the [1.0 release page](https://github.com/eclipse/wakaama/releases/tag/v1.0), will update the README too.

rettichschnidi:
Just added a note to the readme.

rettichschnidi:
@sbernard31 I finally updated the readme. What do you think?

Issue #574, opened 2021-04-02
Automatic code formatting


rettichschnidi:
Historically, the Wakaama code base did not use any kind of code formatting tools. Therefore, every file looks slightly different, making it more hard than necessary to read the code, figuring out and sticking to the predominant style.

Therefore, for C code I propose to pick clang-format (version 10 for now), using a code style close to the one LLVM, the creators of clang-format, use. Foreseeable adaptions might be:
- IndentWidth: 4 (instead of 2)
- ColumnLimit: 120 (instead of 80; current often oversteps the 80 characters boundary)

For CMake files, https://github.com/cheshirekow/cmake_format seems to be the tool to pick.

Acceptance criteria:
- Changes must be done in commits which are limited exclusively to cosmetic/style changes
- Above commit id(s) can be stored in a [--ignore-revs-file](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt) file (supported since Git 2.23).
- GitHub Action in place to prevent merging of non-formatted code
- Tools and exact versions to use defined
- Files in coap/er-coap-13 (and examples/shared/tinydtls) should not be touched
- <s>Code and comments checked for being garbled, exceptions in place where sensible</s> did not format the whole code base

Please note:
- This is issue complements #513

Any opinions on this?

sbernard31:
I guess we definitely need something like this.  :+1: 

About tools to use and format rules, I let the team decide what best fits.

tuve:
I am really in favor of picking a standard like clang-format as suggested, as I do not see any real advantage in relation to the effort to maintain our own standard (or deviation from an existing standard).

qleisan:
I'm in favor of adopting a code standard. No strong opinions on the specifics, what is suggested above works for me. Before any decision is made we should get input from @sbertin-telular also.

sbertin-telular:
I'm also in favor of a code standard with no strong opinions on the specifics. If it can easily be automated and enforced even better.

rettichschnidi:
I tried different clang-format style definitions to approximate the style being used most often. However, always ended up with man thousands of line changes. And while using --ignore-revs-file would work for Git, many 3rd-party-tools (like SonarQube) do not support it, pinning all the code issues on me... :/

Therefore, I implemented a partial solution for this issue: #595

It  enforces the code style on *new* code, which is far less intrusive than formatting all code. Would like to get some feedback on it.

rettichschnidi:
[...]

PR #653, opened 2022-01-30
Modernize CMake

rettichschnidi:
Before this, the build system set many flags and defines globally,
making it hard-to-impossible to build binaries with freely chosen
combinations of defines in the same build tree.

With this new design, compilation flags are bound to their targets,
which is the preferred style since CMake ~3.0.

Pro:
 - Unit tests and compilation checks can cover arbitrary (all!) defines
 - Simplified usage when including Wakaama in a CMake based project:
   - include wakaama.cmake
   - add_executable(your-binary ...)
   - target_compile_definitions(your-binary ...)
   - target_sources_wakaama(your-binary)
 - DTLS enabled example client built next to the plaintext one. This
   will make it easier to integrate both binaries in the same
   integration test run.

Con:
 - Building the example client project requires the tinydtls submodule
   to be checked out. Same goes for the top level "wakaama" project,
   which includes the example client.

As most CMake files had to be changed to implement, I also integrated CMake code formatting enforcement as discussed in #574.


mlasch (approved):
Looks good with `cmake-format` in place.

Issue #663, opened 2022-05-23
Client: Reset context without closing and reconfiguring 


pehrnord:
When running a wakaama client on a connection which might be lost without warning, it is not possible to reset the lwm2m_context in preparation for connection reestablishment. If a connection is reestablished within re-transmission time there will still be items in the transactionList and these will interfere with the (re-)registration process.

Would a "lwm2m_reset" function be a suitable addition for this purpose? Pull request will be added

boaks:
Just for information:

Leshan is able to "match" the client using the peer's "Principal". 
That means: even if a new handshake is executed, the above layers are not affected by that new handshake, as long as the Principal (identity) stay the same. Some years ago, this was discussed over and over in the oma/lwm2m. On argument against that was [RFC 7252 9.1.2](https://datatracker.ietf.org/doc/html/rfc7252#section-9.1.2). A [discussion on the IETF core mailing list](https://mailarchive.ietf.org/arch/msg/core/h8NoXBwt3ttVkB5QZglMoW6Y--Y/) then shows, that this definition has some weakness.

So, I'm not that clear, what "will interfere". 

Also, as least in the past, some DTLS libraries suppert a "DTLS role exchange" as well. That means, your client may not be even not aware, that there is a new DTLS session (at least not without listen to additional library events).

dbarallat:
Based on current test, I agree with @pehrnord: it seems that Leshan is not handling propertly reconfiguration of observations after a client being impacted by a change in its IP address.

Let me add a log, comments:
    - client is a wakaama interfacing with Leshan server
    - I am adding only an "abstract" of client's log to make it easier to follow
    - As a summary:
        a) Battery /3/0/9 changes to 56 but fails to send notify because device is offline
        b) As a result of above, device forces itself going online to send pending notification
        c) Once online:  performs first a Register Update, then sends NOTIFY /3/0/9 to server
        d) Server responds first with an ACK of Register Update and then...
        e) Server responds with a RST to received notification. Under LwM2M, this means server is not recognizing received notification but requests client to stop retransmitting that notificacion
        f) After above, client behaves ok as following /3/0/9 resource updates do not trigger any notification

`...`
`         targetP: 0x4b8f80a0   uri:/3/0/9`
`                watcherP: 0x4b8f80c0    active: 1       update: 0`
`...` 
`[engine]:       *** event battery changed to 56`
`...`
`         targetP: 0x4b8f80a0   uri:/3/0/9`
[...]

PR #664, opened 2022-05-23
Add function to reset lwm2m-context of a client

pehrnord:
Signed-off-by: Pehr Nord <pehr.nord@husqvarnagroup.com>

#663


pehrnord:
Does lwm2m_context.nextMID need to be reset as well?

boaks:
> Does lwm2m_context.nextMID need to be reset as well?

Maybe, you check [RFC7252, Section 4](https://datatracker.ietf.org/doc/html/rfc7252#section-4) about the usage of the MID. Then you may describe the issue you consider in your case and how a reset or not reset will solve that.

In my experience, don't reset it. I only know one issue with MIDs after client restart: you may hit an coap-exchange, which is still kept for deduplication. But the probability of that doesn't change, if you reset the MID in advance. You may just check, if the response (if piggybacked) matches by the MID but not by the token. If that happens after a device restart, you may consider to reset the device MID by a new random value. Or wait for 247s (EXCHANGE_LIFETIME) in order to ensure, that the server's deduplication store is cleaned up.

rettichschnidi:
@pehrnord Will you rework this PR?

PR #666, opened 2022-06-15
server: Make client context available in de-registration callback

mlasch:
Remove the client from server context only after calling the monitoring callback when handling a de-registration. A as result, the client information is still available in the server context during the callback.

Signed-off-by: Marc Lasch <marc.lasch@husqvarnagroup.com>

### DIFF
--- a/core/registration.c
+++ b/core/registration.c
@@ -2018,12 +2018,17 @@ uint8_t  registration_handleRequest(lwm2m_context_t * contextP,
         if (!LWM2M_URI_IS_SET_OBJECT(uriP)) return COAP_400_BAD_REQUEST;
         if (LWM2M_URI_IS_SET_INSTANCE(uriP)) return COAP_400_BAD_REQUEST;
 
-        contextP->clientList = (lwm2m_client_t *)LWM2M_LIST_RM(contextP->clientList, uriP->objectId, &clientP);
-        if (clientP == NULL) return COAP_400_BAD_REQUEST;
-        if (contextP->monitorCallback != NULL)
-        {
-            contextP->monitorCallback(contextP, clientP->internalID, NULL, COAP_202_DELETED, NULL, LWM2M_CONTENT_TEXT, NULL, 0, contextP->monitorUserData);
+        clientP = (lwm2m_client_t *)LWM2M_LIST_FIND(contextP->clientList, uriP->objectId);
+
+        if (clientP == NULL) {
+            return COAP_400_BAD_REQUEST;
         }
+
+        if (contextP->monitorCallback != NULL) {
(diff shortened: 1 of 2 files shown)


core/registration.c
```diff
@@ -803,7 +803,7 @@ static uint8_t prv_register(lwm2m_context_t * contextP,
     dataP->payload = payload;
     dataP->query = query;
     dataP->server = server;
-
```
rettichschnidi, line 806: Please leave out unrelated, cosmetic changes.

core/registration.c
```diff
@@ -2007,23 +2007,26 @@ uint8_t  registration_handleRequest(lwm2m_context_t * contextP,
     }
     break;
 
-    case COAP_DELETE:
-    {
-        lwm2m_client_t * clientP;
+    case COAP_DELETE: {
+        lwm2m_client_t *clientP;
+
+        if (!LWM2M_URI_IS_SET_OBJECT(uriP))
```
rettichschnidi, line 2013: Please always add curly braces.

rettichschnidi (changes requested):
How did this get tested? Any chance for an automated test?

examples/server/lwm2mserver.c
```diff
@@ -974,16 +972,13 @@ static void prv_monitor_callback(lwm2m_context_t *lwm2mH, uint16_t clientID, lwm
...
-        targetP = (lwm2m_client_t *)lwm2m_list_find((lwm2m_list_t *)lwm2mH->clientList, clientID);
-
-        prv_dump_client(targetP);
         break;
 
     default:
         fprintf(stdout, "\r\nMonitor callback called with an unknown status: %d.\r\n", status);
         break;
     }
+    prv_dump_client(clientP);
```
rettichschnidi, line 981: Is it intended to do this even when returning COAP_202_DELETED?

rettichschnidi (changes requested):
Why not just changing the interface of the monitor callback to include a reference to client instead of just the clientID? This way, the client reference *can* be removed from the list before calling the monitor callback.

LukasWoodtli:
LGTM

LukasWoodtli (approved):

examples/server/lwm2mserver.c
mlasch: yes

mlasch:
> Why not just changing the interface of the monitor callback to include a reference to client instead of just the clientID? This way, the client reference _can_ (not sure if sensible) be removed from the list before calling the monitor callback.

This would be an alternative, but it would require quite significant changes in the whole code base. I would prefer to keep it as is for now. Can you open a ticket so this can be discussed in a broader sense?

PR #668, opened 2022-07-03
Gardena/ml/fix uint to text

### LINKED ISSUE #667: warning: absolute value function 'labs' given an argument of type 'int64_t'
I am trying to compile the library for an embedded system based on an ARM Cortex A7.
I have not been able to compile the project yet due to several errors that I have to fix to complete the port.

In particular, I noticed this warning:
image

The `labs` and `llabs` functions are defined as follows:
`long labs (long);`
`long long llabs (long long);`

Where the long type has a size of 4 bytes.

Any suggestions?

mlasch:
Replace `labs()` with `llabs()` to avoid truncation on 32bit architectures. The unit tests now cover the whole `intToText` conversion. This will resolve #667.

Introduces also the compile flag `-Werror=absolute-value` to avoid such issues in the future.


tests/convert_numbers_test.c
```diff
@@ -28,8 +28,18 @@
 #include <inttypes.h>
 #include <float.h>
 
-static const int64_t ints[]={12, -114 , 1 , 134 , 43243 , 0, -215025, INT64_MIN, INT64_MAX};
-static const char* ints_text[] = {"12","-114","1", "134", "43243","0","-215025", "-9223372036854775808", "9223372036854775807"};
+static const int64_t ints[] = {12, -114, 1, 134, 43243, 0, -215025, -4294967297, 4294967297, INT64_MIN, INT64_MAX};
```
rettichschnidi, line 31: Why exactly -4294967297 and 4294967297? Would rather have expected something around INT32_MAX/INT32_MIN (for the value as well as for readability)

core/utils.c
```diff
@@ -218,7 +218,7 @@ size_t utils_intToText(int64_t data,
             string[int64_min_strlen] = '\0';
             return int64_min_strlen;
         }
-        result = utils_uintToText((uint64_t)labs(data), string + 1, length - 1);
+        result = utils_uintToText((uint64_t)llabs(data), string + 1, length - 1);
```
rettichschnidi, line 221: I think the casts to uint64_t are pointless here (and in some other places in the code base). Best case they do nothing, worst case they silence some helpful warnings. How about removing them?

tests/convert_numbers_test.c
mlasch: Important to test are values which overflow int32 and are not exactly `INT64_MIN`, `INT64_MAX`. Looked for better suiting numbers which are more self-explanatory.

rettichschnidi: How about `INT32_MAX + 1` and `INT32_MIN - 1`? Maybe with some casting if need be?

rettichschnidi (approved):

parmi93:
Thanks to both of you!

PR #687, opened 2023-03-02
Disable some compiler warnings from tinydtls

LukasWoodtli:
Not sure if this is a good idea. But the alternative would be to fix tinydtls.

The code of tinydtls is not under our control. It is also not part of Wakaama and is
only used for some examples.
Disabling these warnings lead the focus on warnings emitted from Wakaama code.


rettichschnidi:
@LukasWoodtli Still needed?

PR #700, opened 2023-05-26
fix: avoid example make error make error on macos

L-607:
avoid example make error make error on macos
iShot_2023-05-26_19 54 32


rettichschnidi:
Before we can integrate, please sign the [ECA](https://accounts.eclipse.org/user/eca) so that CI passes.

Issue #701, opened 2023-05-26
[BUG] Memory leak - Bootstrap CoAP temporary payload not deallocated


parmi93:
I noticed a big memory leak caused by the CoAP data that Wakaama receives during the bootstrap phase which are not deallocated.

Specifically, on line 203 a struct `lwm2m_block_data_t` and the string `blockData->identifier.uri` are allocated and the `lwm2m_block_data_t` is added to the `pBlockDataHead` list, which (AFAIK) holds all the temporary CoAP payloads that the bootstrap server is sending to the client (Wakaama), but this can be considered a minor issue because we are talking about a few bytes (about 50 bytes).
The biggest memory leak is caused by the buffer allocated on line 213 which contains the actual CoAP payload sent by the bootstrap server, in my case it reached a size of almost 2KB (contains x509 certificates).
https://github.com/eclipse/wakaama/blob/94506859ad202b85c4d38e3d6f946f4e884540e3/coap/block.c#L201-L216  
  
 
I expected this block of data to be deleted by calling `prv_deleteBootstrapServer(...)`
https://github.com/eclipse/wakaama/blob/94506859ad202b85c4d38e3d6f946f4e884540e3/core/liblwm2m.c#L125-L134

So I modified this function as follows:
https://github.com/parmi93/wakaama/blob/ceddf224ea787c897ac3cb5bd4ed9defc2bd4c93/core/liblwm2m.c#L125-L147
```c
static void prv_deleteBootstrapServer(lwm2m_server_t * serverP, void *userData)
{
    LOG("Entering");
    // TODO should we parse transaction and observation to remove the ones related to this server ?
    if (serverP->sessionH != NULL)
    {
        lwm2m_close_connection(serverP->sessionH, userData);
    }
    if (NULL != serverP->location)
    {
        lwm2m_free(serverP->location);
    }

    while(serverP->blockData != NULL)
    {
        lwm2m_block_data_t * targetP;
        targetP = serverP->blockData;
        serverP->blockData = serverP->blockData->next;
        free_block_data(targetP);
    }

    lwm2m_free(serverP);
}
``` 
Basically I copied the [`prv_deleteServer(...)`](https://github.com/parmi93/wakaama/blob/ceddf224ea787c897ac3cb5bd4ed9defc2bd4c93/core/liblwm2m.c#L91) function.
<br>
### Breafly
[...]

PR #702, opened 2023-05-26
Fix memory leak - Bootstrap CoAP temporary payload not deallocated

parmi93:
Fix a memory leak, caused by the temporary CoAP payload data (send by bootstrap server) not deallocated.
See: https://github.com/eclipse/wakaama/issues/701


parmi93:

mlasch:
Closed in favor of https://github.com/eclipse/wakaama/pull/753

PR #703, opened 2023-05-30
Move cmake toplevel includes to subdirectories

mlasch:
Move the `wakaama.cmake` includes to the sub-directories and add the missing include in tests.

This is needed to use `tests/CMakeLists.txt` standalone.

Steps to reproduce:
```
cd tests
cmake -GNinja -B build-tests .
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 3.13)
 
 project(wakaama C)
 
-include(wakaama.cmake)
-
 add_subdirectory(examples)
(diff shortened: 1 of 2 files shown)


rettichschnidi (approved):
👍

Would however change the commit messages to say what the goal of this change is (to be able to use tests/CMakeLists.txt directly) instead of repeating what the code changes already shows ("Move the wakaama.cmake includes to the subdirectories and add the missing include in tests.")

mlasch:
Made the commit messages more descriptive.

Issue #693, opened 2023-04-03
[BUG?] Memory leak if the client fails to register with the server


parmi93:
It took me a while to figure out where this memory leak occurred, but I'm still not entirely sure if it's a Wakaama bug, or my misuse of this library.

This is the [`lwm2m_transaction_t`](https://github.com/eclipse/wakaama/blob/snapshots/2023-03-31/include/liblwm2m.h#L685) struct:
```c
/*
 * LWM2M transaction
 *
 * Adaptation of Erbium's coap_transaction_t
 */

typedef struct _lwm2m_transaction_ lwm2m_transaction_t;

typedef void (*lwm2m_transaction_callback_t) (lwm2m_context_t * contextP, lwm2m_transaction_t * transacP, void * message);

struct _lwm2m_transaction_
{
    lwm2m_transaction_t * next;  // matches lwm2m_list_t::next
    uint16_t              mID;   // matches lwm2m_list_t::id
    void *                peerH;
    uint8_t               ack_received; // indicates, that the ACK was received
    time_t                response_timeout; // timeout to wait for response, if token is used. When 0, use calculated acknowledge timeout.
    uint8_t  retrans_counter;
    time_t   retrans_time;
    void * message;
    uint16_t buffer_len;
    uint8_t * buffer;
    size_t
        payload_len;  // the length of the entire payload, message payload might be smaller in case of a block1 transfer
    uint8_t *payload; // carries the entire payload across multiple transactions in case of a block 1 transfer
    lwm2m_transaction_callback_t callback;
    void * userData;
};
```
I noticed that the `userData` field is used/allocated by the [`prv_register`](https://github.com/eclipse/wakaama/blob/snapshots/2023-03-31/core/registration.c#L725) function, which is called when the client needs to be registered on the server, in this particular case `userData` is a pointer to another struct ([`registration_data_t`](https://github.com/eclipse/wakaama/blob/snapshots/2023-03-31/core/registration.c#L631)) which contains other pointers (`query` and `payload`) also allocated by the same [`prv_register`](https://github.com/eclipse/wakaama/blob/snapshots/2023-03-31/core/registration.c#L725) function.

[...]

PR #696, opened 2023-04-19
Fix memory leak if the client fails to register with the server

parmi93:
Fix memory leak when client register to the server
See: https://github.com/eclipse/wakaama/issues/693

PR #705, opened 2023-05-31
Fix memory leak if the client fails to register with the server

parmi93:
[registration_data_t](https://github.com/eclipse/wakaama/blob/snapshots/2023-03-31/core/registration.c#L631) is not feed if the server does not respond to the connection message or the user decides to quit the (Wakaama) client before the server respond.
See: https://github.com/eclipse/wakaama/issues/693


rettichschnidi:
closing this PR and reopening #696 to keep things simple.

PR #706, opened 2023-05-31
lwm2m_handle_packet trying to free a NULL payload pointer

parmi93:
lwm2m_handle_packet frees the payload on line 644 without taking into consideration that the payload may not exist, for example in case the client has to send a message without payload (which often happens during the bootstrap phase and connection to the LwM2M server).
See: https://github.com/eclipse/wakaama/issues/707


LukasWoodtli:
I don't think that this is necessary. `lwm2m_free` just calls the `free` function from the standard library. And it's absolutely valid to call `free` with `NULL`.

rettichschnidi:
Closing due to the reasons given by @LukasWoodtli.

PR #709, opened 2023-06-09
Add CBOR support

mlasch:
Add CBOR support to LwM2M server and client introduced with the LwM2M 1.1 standard. This PR is based on work from @sbertin-telular with changes to work with the current build system. Thanks for the contribution.

CBOR support can be enabled by compiling the client with `LWM2M_SUPPORT_SENML_CBOR` flag enabled. The server will support SenML-CBOR when compiled with LwM2M 1.1 support.


data/senml_cbor.c
```diff
@@ -139,6 +139,8 @@ static int prv_parseItem(const uint8_t *buffer, size_t bufferLen, senml_record_t
             bvSeen = true;
             memcpy(baseValue, &data, sizeof(lwm2m_data_t));
             /* Convert explicit 0 to implicit 0 */
+            _Pragma("GCC diagnostic push");
+            _Pragma("GCC diagnostic ignored \"-Wfloat-equal\"");
```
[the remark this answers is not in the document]
sbertin-telular: I really don't like the compiler specific _Pragma. Not everyone uses gcc.

data/cbor.c
```diff
@@ -14,12 +14,14 @@
  *    Scott Bertin, AMETEK, Inc. - Please refer to git log
  *
  *******************************************************************************/
-
+#define _DEFAULT_SOURCE
+#include <endian.h>
```
sbertin-telular, line 18: endian.h is glibc specific and may not exist with other compilers or libraries.

data/senml_cbor.c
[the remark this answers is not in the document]
sbertin-telular: I think there are just 3 equality checks in there outside of the examples and tests. I'm more concerned with the base library than those.

1. Check for NaN which could be replaced by isnan(), I was just targeting C90 when I wrote it.
2. Check whether or not anything would be lost converting double to float (and back). If not, then it can be encoded single precision. Another method would probably need to break it down into the exponent and mantissa and check those. That may need to make assumptions about the underlying hardware representation, although that assumption is probably already being made.
3. Explicitly check for 0. Something like fabsf(a - b) < FLT_EPSILON would involve extra floating point computations and classify anything less than FLT_EPSILON as 0, which would be incorrect. Again, checks on the mantissa and exponent could be done, but be careful that -0 and +0 are treated the same. That would also likely be more computation than a simple compare.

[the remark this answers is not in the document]
LukasWoodtli: I think we could use `fpclassify` since we rely on C11.

tests/cbor_tests.c
```diff
@@ -0,0 +1,1550 @@
...
+((M) == LWM2M_CONTENT_OPAQUE ? "Opaque" :          \
+((M) == LWM2M_CONTENT_TLV ? "TLV" :                \
+((M) == LWM2M_CONTENT_JSON ? "JSON" :              \
+((M) == LWM2M_CONTENT_SENML_JSON ? "SenML JSON" :  \
+((M) == LWM2M_CONTENT_CBOR ? "CBOR" :              \
+((M) == LWM2M_CONTENT_SENML_CBOR ? "SenML CBOR" :  \
+"Unknown"))))))))
+#endif
+
+#if 0
```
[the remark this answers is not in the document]
mlasch: I just imported the code and my IDE conveniently hides the section. I will remove it.

data/senml_cbor.c
```diff
@@ -562,9 +565,9 @@ static int prv_serializeValue(const lwm2m_data_t *tlvP, uint8_t *buffer, size_t
     return head + res;
 }
 
-static int prv_serializeTlv(const lwm2m_data_t *tlvP, const uint8_t *baseUriStr, size_t baseUriLen,
-                            uri_depth_t baseLevel, const uint8_t *parentUriStr, size_t parentUriLen, uri_depth_t level,
-                            bool *baseNameOutput, uint8_t *buffer, size_t bufferLen) {
+static int prv_serializeCbor(const lwm2m_data_t *tlvP, const uint8_t *baseUriStr, size_t baseUriLen,
```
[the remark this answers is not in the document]
mlasch: I wasn't aware that the TLV term is also used for the internal data representation `lwm2m_data_t`. I will change it back.

core/internals.h
```diff
@@ -107,14 +107,41 @@
...
+                              : ((M) == LWM2M_CONTENT_JSON                                                             \
+                                     ? "LWM2M_CONTENT_JSON"                                                            \
+                                     : ((M) == LWM2M_CONTENT_SENML_JSON                                                \
+                                            ? "LWM2M_CONTENT_SENML_JSON"                                               \
+                                            : ((M) == LWM2M_CONTENT_CBOR                                               \
+                                                   ? "LWM2M_CONTENT_CBOR"                                              \
+                                                   : ((M) == LWM2M_CONTENT_SENML_CBOR ? "LWM2M_CONTENT_SENML_CBOR"     \
+                                                                                      : "Unknown"))))))))
+
+/* clang-format off */
```
rettichschnidi, line 128: Move `/* clang-format off */` up, so that this uglification does not happen.

core/internals.h
```diff
@@ -288,6 +320,32 @@ typedef enum
     LWM2M_REQUEST_TYPE_DELETE_ALL
 } lwm2m_request_type_t;
 
+#ifdef LWM2M_SUPPORT_SENML_CBOR
+typedef enum {
```
rettichschnidi, line 324: Maybe referencing a relevant RFC or the like?

rettichschnidi (changes requested):

core/internals.h
mlasch: Done

core/internals.h
mlasch: CBOR is specified in an Internet Standard RFC. https://www.rfc-editor.org/rfc/rfc8949.html

core/internals.h
```diff
@@ -378,6 +430,34 @@ int senml_json_parse(const lwm2m_uri_t * uriP, const uint8_t * buffer, size_t bu
...
+#if defined(LWM2M_SUPPORT_JSON) || defined(LWM2M_SUPPORT_SENML_JSON) || defined(LWM2M_SUPPORT_SENML_CBOR)
+int senml_convert_records(const lwm2m_uri_t *uriP, senml_record_t *recordArray, int numRecords,
+                          senml_convertValue convertValue, lwm2m_data_t **dataP);
+lwm2m_data_t *senml_extendData(lwm2m_data_t *parentP, lwm2m_data_type_t type, uint16_t id);
+int senml_dataStrip(int size, lwm2m_data_t *dataP, lwm2m_data_t **resultP);
+lwm2m_data_t *senml_findDataItem(lwm2m_data_t *listP, size_t count, uint16_t id);
+uri_depth_t senml_decreaseLevel(uri_depth_t level);
+int senml_findAndCheckData(const lwm2m_uri_t *uriP, uri_depth_t baseLevel, size_t size, const lwm2m_data_t *tlvP,
+                           lwm2m_data_t **targetP, uri_depth_t *targetLevelP);
+#endif
```
rettichschnidi, line 459: It is not really done at all in Wakaama, but having (at least) the return values documented would be reeeeeeealy nice. How is success encoded? >0? How an error? <0?

tests/cbor_tests.c
```diff
@@ -0,0 +1,1158 @@
...
+                              ? "TLV"                                                                                  \
+                              : ((M) == LWM2M_CONTENT_JSON                                                             \
+                                     ? "JSON"                                                                          \
+                                     : ((M) == LWM2M_CONTENT_SENML_JSON                                                \
+                                            ? "SenML JSON"                                                             \
+                                            : ((M) == LWM2M_CONTENT_CBOR                                               \
+                                                   ? "CBOR"                                                            \
+                                                   : ((M) == LWM2M_CONTENT_SENML_CBOR ? "SenML CBOR"                   \
+                                                                                      : "Unknown"))))))))
```
rettichschnidi, line 55: Looks better formatted by hand.

tests/senml_cbor_tests.c
```diff
@@ -0,0 +1,836 @@
...
+                : ((M) == LWM2M_CONTENT_OPAQUE                                                                         \
+                       ? "Opaque"                                                                                      \
+                       : ((M) == LWM2M_CONTENT_TLV                                                                     \
+                              ? "TLV"                                                                                  \
+                              : ((M) == LWM2M_CONTENT_JSON                                                             \
+                                     ? "JSON"                                                                          \
+                                     : ((M) == LWM2M_CONTENT_SENML_JSON                                                \
+                                            ? "SenML JSON"                                                             \
+                                            : ((M) == LWM2M_CONTENT_SENML_CBOR ? "SenML CBOR" : "Unknown")))))))
```
rettichschnidi, line 50: Looks better formatted by hand.

rettichschnidi (changes requested):

core/internals.h
rettichschnidi: Having a link to this would be really nice anway. But maybe from within `data/cbor.c`?

mlasch: I'm not sure whether the code is generally complaint with rfc8949 or [rfc7049](https://datatracker.ietf.org/doc/html/rfc7049).

sbertin-telular: OMA-TS-LightweightM2M_Core-V1_2-20201110-A references [rfc8428](https://tools.ietf.org/html/rfc8428.txt) for SenML JSON and CBOR.

sbertin-telular:
I'm not involved in any projects using LWM2M anymore, so don't wait for my review. My only objection remains the use of the gcc specific _Pragma.

core/internals.h
sbernard31: Sorry, If missunderstood. 

But since LWM2M v1.1 there is 2 content format which are using CBOR.


Data Format | IANA Media Type | Numeric Content-Formats [CoAP]
-- | -- | --
CBOR | application/cbor | 60
SenML CBOR | application/senml+cbor | 112

(source : [LWM2M-v1.1.1@core§Table: 7.4.-2 Additional standardized Media Types supported in LwM2M TS 1.1](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Core-V1_1_1-20190617-A.html#Table-74-2-Additional-standardized-Media-Types-supported-in-LwM2M-TS-11))

**Usage:**
- First one is only used for singular resource (and resource instance) like Text format.
- Second one is used for everything. (like old JSON format)

**RFC:**
- First one is based on [RFC 7049.](https://www.rfc-editor.org/rfc/rfc7049.html) in LWM2M v1.1.1 but this RFC was obsoleted by [RFC 8949](https://www.rfc-editor.org/rfc/rfc8949). LWM2M v1.2.1 refers RFC 8949. In LWM2M v1.1.1 there very few information about this format, in LWM2M v1.2.1 this is a bit clearer.

- Second one is based on [RFC 8428](https://www.rfc-editor.org/rfc/rfc8428) especially [§6 - CBOR Representation (application/senml+cbor)](https://www.rfc-editor.org/rfc/rfc8428#section-6)  (:point_up: which is based on RFC 7049 which is obsoleted by RFC 8949 :sweat_smile:)

**About RFC 7049  obsoleted by RFC 8949**

RFC 8949 says : 
> This document obsoletes RFC 7049, providing editorial improvements, new details, and errata fixes while keeping full compatibility with the interchange format of RFC 7049. It does not create a new version of the format.

So I guess it is worth to target RFC 8949 even for LWM2M v1.1.1 and SENML-CBOR.

LukasWoodtli (approved):

rettichschnidi (approved):
Not tested/used the code, but looks sane enough to me.

sbertin-telular (approved):

data/senml_cbor.c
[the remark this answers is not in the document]
mlasch: > I think we could use fpclassify since we rely on C11.

Will do this in a separate PR. There are already some Pragma in the code to mute floating point comparison compiler warnings. They can be removed as well.

Issue #698, opened 2023-05-03
[BUG] Opaque value decoded in base64 format instead of base64 URL format


parmi93:
I was thinking that was a Leshan bootstrap demo server bug, but it seems to be a problem with the Wakaama library.
https://github.com/eclipse/leshan/issues/1444
https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/553

According to the [RFC8428§4.3- SenML Labels](https://www.rfc-editor.org/rfc/rfc8428#section-4.3) the opaque value over SenML JSON should be encode in base64 URL (no padding) format, but Wakaama lib is trying to decode as base64 (with padding) format.

I noticed this issue because I'm failing the bootstrap phase because of that.

Has anyone else noticed the same problem?

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #710, opened 2023-06-13
Fix - Changed encoding from Base64 (with padding) to Base64URL (no padding) for SenML JSON format

parmi93:
According to the SenML JSON specification defined at [RFC8428§4.3- SenML Labels](https://www.rfc-editor.org/rfc/rfc8428#section-4.3) the Base64 URL safe alphabet must be used (without padding) for Data Value `vd`:

> (*) Data Value is a base64-encoded string with the URL-safe alphabet as defined in [Section 5 of [RFC4648]](https://www.rfc-editor.org/rfc/rfc4648#section-5), with padding omitted. (In CBOR, the octets in the Data Value are encoded using a definite-length byte string, major type 2.)

So, encoding and decoding has been changed from Base64 (with padding) to Base64Url (without padding) for data over SenML JSON format.

Note that the behavior of the code in `data/json.c` has **not** changed, data in JSON format will continue to be encoded in Base64 (with padding), see [eclipse/leshan#1444 (comment)](https://github.com/eclipse/leshan/issues/1444#issuecomment-1588852832).

New tests has been added in order to test the Base64Url (without padding) encode and decode functions, and the previous tests have been modified by adding a new entry  to be tested.

Fix: https://github.com/eclipse/wakaama/issues/698
also see: https://github.com/OpenMobileAlliance/OMA_LwM2M_for_Developers/issues/553

### DIFF
--- a/core/internals.h
+++ b/core/internals.h
@@ -446,10 +446,10 @@ int utils_textToObjLink(const uint8_t * buffer,
                         uint16_t * objectId,
                         uint16_t * objectInstanceId);
 void utils_copyValue(void * dst, const void * src, size_t len);
-size_t utils_base64GetSize(size_t dataLen);
-size_t utils_base64Encode(const uint8_t * dataP, size_t dataLen, uint8_t * bufferP, size_t bufferLen);
+size_t utils_base64GetSize(size_t dataLen, bool withPadding);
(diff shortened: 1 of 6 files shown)


parmi93:

parmi93:
Removed the history of commits, in order to be able to resolve errors related to commit messages that did not respect the rules.

PR #711, opened 2023-06-13
Made srand(...) and rand() platform independent

parmi93:
The `srand(...)` and `rand()` functions are not available on all embedded platforms, so they were replaced with `lwm2m_srand(...)` and `lwm2m_rand()` to make these calls platform independent.

### DIFF
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -68,7 +68,7 @@ lwm2m_context_t * lwm2m_init(void * userData)
     {
         memset(contextP, 0, sizeof(lwm2m_context_t));
         contextP->userData = userData;
-        srand((int)lwm2m_gettime());
+        srand(lwm2m_seed());
         contextP->nextMID = rand();
     }
 
--- a/examples/shared/platform.c
+++ b/examples/shared/platform.c
@@ -67,6 +67,16 @@ time_t lwm2m_gettime(void)
     return time(NULL);
 }
 
+int lwm2m_seed(void) {
+    /*
+     * Return a seed for random number generation, the seed must be a
+     * different number at every boot and unpredictable, time(NULL) may not be
+     * a reliable source as a seed.
+     * See: https://github.com/eclipse/wakaama/pull/711
+     */
+    return time(NULL);
+}
+
 void lwm2m_printf(const char * format, ...)
 {
     va_list ap;
--- a/include/liblwm2m.h
+++ b/include/liblwm2m.h
@@ -118,6 +118,8 @@ int lwm2m_strcasecmp(const char * str1, const char * str2);
 // In case of error, this must return a negative value.
 // Per POSIX specifications, time_t is a signed integer.
 time_t lwm2m_gettime(void);
+// Get a seed (which must not repeat when the device reboots) for generating a random number
+int lwm2m_seed(void);
 
 #ifdef LWM2M_WITH_LOGS
 // Same usage as C89 printf()


parmi93:

parmi93:
I noticed that the messages of all previous commits are also checked (which kept failing), so I deleted the previous commits, leaving only the last commit in the history.
Can you please start compliance checks again? I hope this works

mlasch:
Can you share on which (embedded-)platform you are running Wakaama? It looks like you still include `stdlib.h`, so I assume some kind of standard C library is available. srand() and rand() are C89/C90 standard (ISO/IEC 9899:1990) and newer.

I'm wondering what's the best way how to deal with exotic platforms in general. I don't think it's a good practice to re-implement and wrap standard C functions. In my opinion we should assume that we can rely at least on the ISO C standard library (and POSIX?). However the project does not state any minimal requirements yet. Might be worth a discussion #494.

If (s)rand() is not available on your platform you might implement your own rand() function without the wrapper? I guess the same goes for malloc(), free() etc..

LukasWoodtli:
Even if I like the idea to build platform independent software. I would try to reduce the customization possibilities in the code. 

I would prefer to leverage the possibilities of the toolchain (compiler, linker...) or the OS (loader) for such customization. These tools have all we need and it reduces some unneccessary complexety of our code.

If you want to use your custom `srand()` funciton you can just implement it in one of your C files.

```
void srand (unsigned int seed) {
    (void)seed;
    fprintf(stdout, "Hello from srand");
}

int main(int argc, char *argv[])
{
// ...
}
```
The linker will pick your implementation instead the one from the standard library.

parmi93:
I understand your point of view.

But I think the problem is still there.
This line of code is implicitly assuming that the system clock is synchronized:
https://github.com/eclipse/wakaama/blob/514659414c792f8c252782af63551b6d09704a7b/core/liblwm2m.c#L71
I don't think Wakaama can take this for granted, because even under POSIX (Linux) system this might not be true (e g. backup battery is flat).
At most we can take for granted that the timestamp returned by `lwm2m_time()` is incremented by the system every second.

Perhaps it would be better to replace with:
```c
srand((int)lwm2m_seed());
```
And let the user define the `lwm2m_seed()` function in the `platform.c` file?

LukasWoodtli:
I see your point. So probably we need a possibility to customise `srand()`.
Another idea would be to provide a second argument to `lwm2m_init` with the seed. But I don’t know if that is feasible.

mlasch:
@parmi93 all valid points, thanks for clarifying. I like your proposed solution to have a custom seed function.

> Perhaps it would be better to replace with:
> ```c
> srand((int)lwm2m_seed());
> ```
> And let the user define the lwm2m_seed() function in the platform.c file?

mlasch:
I would suggest to keep `time(NULL)` as a default seed source due to the lack of a better option. Using `lwm2m_seed()`  instead of `lwm2m_gettime` is much clearer.

parmi93:
Commits squashed and commit message subject line changed.

core/liblwm2m.c
```diff
@@ -68,7 +68,7 @@ lwm2m_context_t * lwm2m_init(void * userData)
     {
         memset(contextP, 0, sizeof(lwm2m_context_t));
         contextP->userData = userData;
-        srand((int)lwm2m_gettime());
+        srand((int)lwm2m_seed());
```
mlasch, line 71: The type cast to `int` is not needed anymore, since `lwm2m_seed()` already returns an `int`.

parmi93: Fixed

mlasch:
The changes look good. Now the CI is still complaining about formatting. You can use `clang-format` to auto-format your changes. e.g.
```
git clang-format HEAD~1
```

parmi93:
Code formatted with `clang-format`.

PR #712, opened 2023-06-13
Renamed uri_toString(...) to lwm2m_uriToString(...) in order to make it public

parmi93:
I often had to convert a `lwm2m_uri_t` object to string, so I made the `uri_toString(...)` function public and renamed it to `lwm2m_uriToString(...)`


parmi93:

mlasch:
Looks good to me.

PR #713, opened 2023-06-14
lwm2m_remove_object(...) return removed object

parmi93:
Small change to the `lwm2m_remove_object(...)` function so that it returns the pointer to the removed object.


parmi93:

parmi93:
Removed the history of commits, in order to be able to resolve errors related to commit messages that did not respect the rules.

rettichschnidi:
When exactly will this be useful? Would be great to have this information also in the commit message.

PR #718, opened 2023-06-29
doc: Specify C standard and library requirements

LukasWoodtli:
Specify the use of C11 and requirements regarding the standard library and POSIX (Linux).

AFAIK these requirement have been defined elsewhere. It's now just visibly documented here.


mlasch:
I'm afraid we cannot solely rely on Linux. There is also a discussion here: #494. But it is a good starting point and important to have the result of the discussion fixed in the readme.

LukasWoodtli:
I just realised that Wakaama was successfully ported to different platforms. 
I will remove the part about POSIX/Linux.

LukasWoodtli:
The information I added to the readme is not really useful. Abandoning...

PR #723, opened 2023-07-27
examples: use getline() to read commands

raiden00pl:
examples: use getline() to read commands

The previous approach works only if the terminal is operating in cannonical mode (input available line-by-line), which is not always true. getline() ensures that commands are always read correctly.

With this change commands work properly on Apache NuttX

### DIFF
--- a/examples/client/lwm2mclient.c
+++ b/examples/client/lwm2mclient.c
@@ -1407,15 +1407,18 @@ int main(int argc, char *argv[])
              */
             else if (FD_ISSET(STDIN_FILENO, &readfds))
             {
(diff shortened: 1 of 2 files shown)


mlasch:
Hi @raiden00pl, thanks for your contribution. Nice to see on which platforms Wakaama is integrated.

There are no regression tests for Apache NuttX yet, so we can not guarantee that we won't break NuttX with future changes.

raiden00pl:
No problem with that. NuttX tries to be compatible with Linux as much as possible for a system dedicated to embedded applications, so it shouldn't be that easy to break NuttX support ;)

Issue #491, opened 2020-11-04
Add community "classic" documentation. 📘


sbernard31:
Maybe wakaama need some common community documentation.

Regarding [Github community](https://github.com/eclipse/wakaama/community), it missed not so much : 
- [x] Description
- [x] README
- [x] Code of conduct
- [x] Contributing
- [ ] License
- [ ] Issue templates
- [ ] Pull request template
- [ ] Repository admins accept content reports 

(Maybe even already checked task need to be reviewed)

We should also add more information about : [Security](https://github.com/eclipse/wakaama/security)

Wiki can also be used to add some information (like roadmap, faq ...)

Even if this is not perfect you could have a look to Leshan project as inspiration.  (or any other eclipse project or popular open source project)
:warning: About Leshan, all documentation is in wiki which is not a so good idea because often documentation and product as same lifecycle, so maybe this is a better idea to store documentation as markdown file with the code.

sbernard31:
I changed the License file name this way github is happy (see commit https://github.com/eclipse/wakaama/commit/2c126779844a0b4086fb7600135ede492854227f)

sbernard31:
I'm not sure what is the purpose of :https://github.com/eclipse/wakaama/blob/master/TODO ? :thinking:

sbertin-telular:
> I'm not sure what is the purpose of :https://github.com/eclipse/wakaama/blob/master/TODO ? 🤔

I'm not sure I've ever looked at that. It is definitely out of date, but does have some good ideas. The following have definitely been implemented, but some of the others might need more investigation.
- JSON support
- Support for Object Link datatype
- Discover method support

sbernard31:
Maybe we can create corresponding issue for good idea, then we remove this file ?
Or you think we should keep it ?

sbertin-telular:
I think creating issues for the good ideas and removing the file would be best. It is too easy to overlook the file.

sbernard31:
I let you create issues (and maybe labelled it correctly if needed) as you know better than me what are the good idea.
Please let me know when it's done and I will remove the file. :pray:

sbernard31:
I added more tag to the project : `iot`, `coap`, `eclipseiot`, `internet-of-things`, `device-management`.
I slightly changed the project description.
I hide the **Package** panel (just below **Releases**) as we don't use it.

sbernard31:
I enabled the [Wiki](https://github.com/eclipse/wakaama/wiki) tab for this project.
[...]

PR #725, opened 2023-08-09
license: Rename license for Github to detect it

mlasch:
Rename license for Github to detect it, similar to Leshan. #491


sbernard31:
I'm not sure it works to use `EDL-1` suffix :thinking:,  did you check on your personal forked repository ?

It works if you can see : BSD-3-Clause, EPL-2.0 licenses found  on the right of the github home page under tags.
Like in Leshan : https://github.com/eclipse-leshan/leshan

Note that EDL-1 license is exactly a BSD-3-Clause. (Leshan is also licensed under epl 2 and edl 1)
Why edl-1 is not supported ? I asked about it long time ago : https://lists.spdx.org/g/Spdx-legal/topic/22080928#2105
So it's expected that GitHub see edl1 as BSD-3 license.

mlasch:
> I'm not sure it works to use EDL-1 suffix 🤔, did you check on your personal forked repository ?

The licenses will show up in Github as "Unknown, EPL-2.0 licenses found ". At least the EDL-1 is recognized as a license file, even though unknown. 

> Note that EDL-1 license is exactly a BSD-3-Clause.

Not sure I understand you correctly. Can we just replace EDL-1 with BSD-3 in Wakaama (same as with Leshan), or should we keep EDL-1 and accept that Github does not show it correctly? Both would be fine for me.

sbernard31:
> Not sure I understand you correctly. Can we just replace EDL-1 with BSD-3 in Wakaama 

I think you can replace EDL-1 with BSD-3. 
@waynebeaton could you confirm that it's OK to replace `EDL-1` suffix by `BSD-3` suffix and so github is able to show `BSD-3-Clause` instance of `Unknown` ?

boaks:
> I think you can replace EDL-1 with BSD-3.

At least that was the advice for Californium.

sbernard31 (approved):

waynebeaton:
Yes. In fact, the [documentation generator](https://www.eclipse.org/projects/tools/documentation.php?id=iot.wakaama#copyright) recommends exactly this.

Use `BSD-3-Clause`  if you can. While there's all kinds of standards, I find that there is a strong preference to using the SPDX code for the name of the file with an extension that describes the format, e.g., `BSD-3-Clause.md` or similar. This is not a requirement... just an observation.

FWIW, the BSD licenses are referred to as "template" licenses. When you use them, you change certain parts of the text, but they're basically still just the BSD license. So the EDL is just our appliaction of the BSD. From a license management perspective, it's way easier for everybody if we treat all of the various applications of the BSD uniformly.

PR #726, opened 2023-08-17
Allow registration message without Resource Type

LukasWoodtli:
Registration messages without `rt="oma.lwm2m"` should be accepted as valid request.


mlasch:
Did some tests, looks good to me.

mlasch (approved):

PR #727, opened 2023-09-21
Add test executables for different configurations

LukasWoodtli:
The same tests are compiled and ran with different defines to make sure that combinations of different configurations work properly.

Some combinations of defines are added. This can be extended to include other interesting configuration combinations.

### DIFF
--- a/.cmake-format
+++ b/.cmake-format
@@ -15,3 +15,13 @@ with section("format"):
 with section("markup"):
   # What character to use for bulleted lists
   bullet_char = '-'
+
+additional_commands = {
+  "add_test_variant": {
+    "kwargs": {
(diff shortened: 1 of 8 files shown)


.cmake-format
```diff
@@ -15,3 +15,13 @@ with section("format"):
...
+
+additional_commands = {
+  "add_test_variant": {
+    "kwargs": {
+      "TARGET_NAME": '*',
+      "SOURCE_FILES": '*',
+      "COMPILE_DEFINITIONS": '*',
+    }
+  }
+}
```
mlasch, line 27: The newline has done missing here.

mlasch (approved):
Nice addition to support automated testing and coverage.

PR #738, opened 2023-11-20
ci: Fix building and running with sanitizers

LukasWoodtli:
The sanitizers need to be activated at configuration time of the CMake project. Not when executing the build command.


rettichschnidi (approved):
Great catch, thanks!

mlasch (approved):

PR #742, opened 2023-11-27
ci: Fix SonarCloud

LukasWoodtli:
The output of the sonar build wrapper was
overwritten when the tests where ran with
the wrapper. The wrapper is needed during
the build. But not for running the tests.


mlasch (approved):
Thanks a lot for fixing.

PR #746, opened 2024-01-03
Testing the response for a simple registration message

LukasWoodtli:
Add a simple test for handling a registration message.

Some testing infrastructure is added that we can assert the reponse sent by Wakaama is as expected.

### DIFF
--- a/tests/helper/connection.c
+++ b/tests/helper/connection.c
@@ -1,0 +1,52 @@
+/*******************************************************************************
+ *
+ * Copyright (c) 2024 GARDENA GmbH
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v1.0 which accompany this distribution.
+ *
+ * The Eclipse Public License is available at
+ *    http://www.eclipse.org/legal/epl-v20.html
+ * The Eclipse Distribution License is available at
+ *    http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * Contributors:
+ *   Lukas Woodtli, GARDENA GmbH - Please refer to git log
+ *
+ *******************************************************************************/
+
+#include <assert.h>
+#include <stdbool.h>
+#include <string.h>
+
+#include "connection.h"
+#include "liblwm2m.h"
+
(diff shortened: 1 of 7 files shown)


tests/helper/connection.c
```diff
@@ -0,0 +1,52 @@
...
+ *******************************************************************************/
+
+#include <assert.h>
+#include <stdbool.h>
+#include <string.h>
+
+#include "connection.h"
+#include "liblwm2m.h"
+
+#define SEND_BUFFER_MAX_LEN 100
```
mlasch, line 26: To me the buffer size seems very small. Max packet size when using block transfer is 1024 + header. If would rather put 2048 here.

tests/registration_tests.c
```diff
@@ -17,12 +17,75 @@
...
+    coap_free_header(&packet);
+
+    /* act */
+    lwm2m_context_t *const server_ctx = lwm2m_init(NULL);
+    lwm2m_handle_packet(server_ctx, reg_coap_msg, msg_size, NULL);
+    lwm2m_close(server_ctx);
+
+    /* assert */
+    size_t send_buffer_len;
+    uint8_t *send_buffer = test_get_send_buffer(&send_buffer_len);
```
mlasch, line 69: If I understand this correctly this buffer then contains the registration response? At first this was not clear to me. Maybe change the variable name to stress that it is the registration response we are checking here. Send buffer is a bit misleading in my opinion.

tests/registration_tests.c
```diff
@@ -17,12 +17,75 @@
...
+#include "er-coap-13/er-coap-13.h"
 #include "internals.h"
 #include "liblwm2m.h"
 #include "tests.h"
 
+#include <string.h>
+
 #ifdef LWM2M_SERVER_MODE
 
+static void create_test_registration_message(coap_packet_t *packet) {
```
mlasch, line 30: This function sounds very generic but creates a very specific message for the test below. Why not having these lines directly in the test function? If we add more tests we would need a new function anyway.

LukasWoodtli: My idea is to have, in the future, multiple helper functions that provide some testing messages for different tests.
Having this in a separate function gives it a descriptive name and make the test function a bit smaller.

mlasch (approved):

PR #749, opened 2024-01-29
ci: Fix SonarCloud Scan with outdated Java version

mlasch:
SonarCloud Scan does no longer support Java 11. Use `sonarsource/sonarcloud-github-c-cpp` action to install sonarcloud and the build-wrapper.


rettichschnidi:
> Use `sonarsource/sonarcloud-github-c-cpp` action to install sonarcloud and the build-wrapper.

Would mention that this is offered by the original vendor, not some 3rd-party action as before.

.github/workflows/sonarcloud-scan.yaml
```diff
@@ -0,0 +1,53 @@
...
+      with:
+        submodules: true
+        fetch-depth: 0
+
+    - name: Install dependencies from APT repository
+      run:  |
+        sudo apt-get update
+        sudo apt-get install gcovr libcunit1-dev wget unzip
+
+    - name: Install CMake
```
rettichschnidi, line 21: Would remove as you did in #750.

rettichschnidi (approved):

PR #753, opened 2024-02-08
Fix memory leak of bootstrap CoAP payload

mlasch:
Fix a memory leak, caused by the temporary CoAP payload data (send by bootstrap server) not deallocated.

See: https://github.com/eclipse/wakaama/issues/701

This is a rebase of https://github.com/eclipse/wakaama/pull/702, adjusted code format as well.

### DIFF
--- a/core/liblwm2m.c
+++ b/core/liblwm2m.c
@@ -124,12 +124,21 @@ static void prv_deleteServerList(lwm2m_context_t * context)
 
 static void prv_deleteBootstrapServer(lwm2m_server_t * serverP, void *userData)
 {
-    // TODO should we free location as in prv_deleteServer ?
+    LOG("Entering");
     // TODO should we parse transaction and observation to remove the ones related to this server ?
-    if (serverP->sessionH != NULL)
-    {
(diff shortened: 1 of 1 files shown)


core/liblwm2m.c
```diff
@@ -124,12 +124,23 @@ static void prv_deleteServerList(lwm2m_context_t * context)
...
 {
-    // TODO should we free location as in prv_deleteServer ?
+    LOG("Entering");
     // TODO should we parse transaction and observation to remove the ones related to this server ?
     if (serverP->sessionH != NULL)
     {
-         lwm2m_close_connection(serverP->sessionH, userData);
+        lwm2m_close_connection(serverP->sessionH, userData);
+    }
+    if (NULL != serverP->location) {
```
LukasWoodtli, line 133: This `if` statement is not necessary. It's valid to free a `NULL` pointer in C.

LukasWoodtli (commented):
Looks good

But see comment about checking for `NULL` before calling `lwm2m_free`

LukasWoodtli (approved):
A test would be nice. But otherwise looks good!

PR #755, opened 2024-02-09
ci: Exclude tests and `tinydtls` from `scan-build`

LukasWoodtli:
We don't want to check the third-party code of `tinydtls`. Checking the tests gives a lot of complaints about macros of `CUint` that are not defined in the context of `scan-build`.

### DIFF
--- a/tools/ci/run_ci.sh
+++ b/tools/ci/run_ci.sh
@@ -372,8 +372,13 @@ if [ -n "${OPT_TEST_COVERAGE_REPORT}" ]; then
 fi
 
 if [ -n "${OPT_SCAN_BUILD}" ]; then
+  # Excluding tests as `scan-build` does not know macros of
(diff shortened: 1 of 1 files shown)


tools/ci/run_ci.sh
```diff
@@ -373,7 +373,9 @@ fi
 
 if [ -n "${OPT_SCAN_BUILD}" ]; then
   OPT_WRAPPER_CMD="${OPT_SCAN_BUILD} \
-    -o build-wakaama/clang-static-analyzer"
+    -o build-wakaama/clang-static-analyzer \
+    --exclude tests \
+    --exclude examples/shared/tinydtls"
```
rettichschnidi, line 378: Would leave a comment in the sources on why this folders have been excluded.

rettichschnidi (approved):

mlasch (approved):

PR #756, opened 2024-02-12
Fix invalid content type for registration

LukasWoodtli:
LwM2M registration is only valid with core link format.


mlasch:
According to 1.1.1 specification section _6.2.1. Register Operation_ 

> the payload Media-Type of that registration message MUST be the CoRE Link Format (application/link-format) defined
in [RFC6690]

as a result plain text must not be allowed.

LukasWoodtli:
> According to 1.1.1 specification section _6.2.1. Register Operation_
> 
> > the payload Media-Type of that registration message MUST be the CoRE Link Format (application/link-format) defined
> > in [RFC6690]
> 
> as a result plain text must not be allowed.

Agree! Fixed Wakaama to be standard conformant.

mlasch:
Nice, thanks!

mlasch (approved):

Issue #770, opened 2024-05-02
Client SEND operation does not use fragmentation leading to oversized CoAP packets


sergioGras:
## Description
I encountered an issue where the client SEND operation does not use the existing fragmentation logic implemented for server WRITE operation. This results in the creation of oversized CoAP packet when large payloads are sent, which the server then rejects due to exceeding the maximum packet size limit. 

## Steps to Reproduce
1. Start the server and example client.
2. Add a new resource that exceeds the 2048 byte limit.
3. Attempt to send the resource from the client using the SEND method (lwm2m_send).
4. Observe that the payload is not fragmented and is sent in a single large CoAP packet.

## Expected Behavior
The client SEND operation should fragment large payloads into multiple CoAP packets, similar to how the server-side WRITE operation handles them.

## Possible Fix
The issue can be resolved by utilizing the _transaction_set_payload_ method in the _lwm2m_send_ method, similar to its usage in the server WRITE operation, instead of the currently used _coap_set_payload_.

LukasWoodtli:
Closing because the Wakaama project will be terminated: See issue #883 and [iot.wakaama 2026.05 Termination Review](https://gitlab.eclipse.org/eclipsefdn/emo-team/emo/-/issues/1191)

PR #771, opened 2024-05-03
Fix payload fragmentation in client SEND operation #770

sergioGras:
In this PR, I replaced the _coap_set_payload_ function with _transaction_set_payload_. As a result, the client can now send large data in separate CoAP packets instead of a single large packet that was previously rejected by the server


LukasWoodtli:
@sergioGras Thank you for your contribution!

You need to sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) to provide a PR.

Please also provide a useful commit message.

mlasch:
You pushed a merge commit with your contribution. Can you please squash your commits into a single commit?

Issue #773, opened 2024-05-03
Build Fails if LWM2M_RAW_BLOCK1_REQUESTS set


mmb-davidsmith:
I'm looking at using Wakaama in a firmware project and was trying to get the various knobs set to enable me to build for a low ram environment. In the documentation, it says that the `LWM2M_RAW_BLOCK1_REQUESTS` is intended for this sort of situation and configures the system to pass through each individual packet to the application rather than assembling the data.

When I enable this option and try to run my copy of the `lightclient` I run into the following compile errors:

```
wakaama/core/objects.c:530:20: error: unused variable 'dataP' [-Werror,-Wunused-variable]
    lwm2m_data_t * dataP = NULL;
                   ^
wakaama/core/objects.c:531:9: error: unused variable 'size' [-Werror,-Wunused-variable]
    int size = 0;
        ^
```

Quickly commenting these out since they're not used in the function, the following errors is raised:
```
wakaama/coap/block.c:294:9: error: conflicting types for 'coap_block1_handler'
uint8_t coap_block1_handler(lwm2m_block_data_t **pBlockDataHead,
        ^
wakaama/core/internals.h:483:9: note: previous declaration is here
uint8_t coap_block1_handler(lwm2m_block_data_t ** blockData, const char * uri, uint16_t mid, uint8_t * buffer, size_t length, uint16_t blockSize, uint32_t blockNum, bool blockMore, uint8_t ** outputBuffer, size_t * outputLength);
        ^
```

This is due to a missing `const` keyword on the `uint8_t * buffer`. Once that is changed, I'm able to compile but the samples don't handle the full block transfer and would need some updates to complete a transfer.

I'm just working on the internal processes to get a Contribution signed off to address this and should be able to contribute back.

LukasWoodtli:
@mmb-davidsmith Thank you for your participation. `LWM2M_RAW_BLOCK1_REQUESTS` has not got any attention lately. It would be nice to have some improvements and clean-ups in this area.
Looking forward to your contribution.

mmb-davidsmith:
@LukasWoodtli - I've submitted PR #777, which gets this building again. It needs approval for the workflows to run.

mmb-davidsmith:
Fixed with the pushed commit.

PR #777, opened 2024-05-22
Fix LWM2M_RAW_BLOCK1_REQUESTS compile

mmb-davidsmith:
Addresses the compile issues referenced in #773 when using the `LWM2M_RAW_BLOCK1_REQUESTS` compile flag by:
- removing unused variables
- specifying the missing const keyword.

In addition, to help validate that this doesn't happen in the future, this commit adds an additional configuration option to the client example which enables the `LWM2M_RAW_BLOCK1_REQUESTS` option. This allows for a basic test that compilation succeeds. It does not go further in attempting to add unit tests for this compile option.


LukasWoodtli:
Thank you for your contribution!

LukasWoodtli (approved):

PR #778, opened 2024-05-26
cmake: Fix define for CoAP block size

LukasWoodtli:
The define was missing the `lwm2m` prefix in the CMake project.


mlasch:
I think this one needs a rebase.

PR #784, opened 2024-06-03
Lw/cmake options defines

LukasWoodtli:
Steps towards a modern CMake project.


LukasWoodtli:
This PR is too big. Splitting in smaller ones. First one: #785

PR #785, opened 2024-06-05
CMake options for defining the mode

LukasWoodtli:
Step to a more modern CMake project


README.md
```diff
@@ -93,9 +89,17 @@ Several preprocessor definitions are supported:
    - LWM2M_VERSION_1_0 to support only version 1.0
    Please note: Clients support only the specified version, while servers are backward compatible.
  - LWM2M_RAW_BLOCK1_REQUESTS For low memory client devices where it is not possible to keep a large post or put request in memory to be parsed (typically a firmware write).
-   This option enable each unprocessed block 1 payload to be passed to the application, typically to be stored to a flash memory. 
+   This option enable each unprocessed block 1 payload to be passed to the application, typically to be stored to a flash memory.
  - LWM2M_COAP_DEFAULT_BLOCK_SIZE CoAP block size used by CoAP layer when performing block-wise transfers. Possible values: 16, 32, 64, 128, 256, 512 and 1024. Defaults to 1024.
 
+### Mode
+
+Wakaama supports multiple modes. At least one mode needs to be defined with CMAke cache variables.
```
mlasch, line 97:
```suggestion
Wakaama supports multiple modes. At least one mode needs to be defined with CMake cache variables.
```

tests/integration/conftest.py
```diff
@@ -6,6 +6,10 @@
 import pytest
 
 
+BASE_PATH = str(Path(__file__).parent.absolute())
+REPO_BASE_PATH = BASE_PATH + "/../../"
```
mlasch, line 10: Why not use the Path object all the way through?

tests/integration/conftest.py
```diff
@@ -67,14 +71,13 @@ class Lwm2mServer(HelperBase):
     """Server subclass of HelperBase"""
 
     def __init__(self, arguments="", timeout=3, encoding="utf8"):
-        base_path = str(Path(__file__).parent.absolute())
-        self.pexpectobj = pexpect.spawn(base_path +
-                                        "/../../build-wakaama/examples/server/lwm2mserver " +
+        self.pexpectobj = pexpect.spawn(REPO_BASE_PATH +
```
mlasch, line 74: If `REPO_BASE_PATH` would be an object of Path, this could be more pythonic.

mlasch (approved):
Looks good. But I have some minor comments.

tests/integration/conftest.py
LukasWoodtli: `pexpect` needs a string as argument. I would also prefer a more pythonic way. But `pexpect` is not really pythonic.

tests/integration/conftest.py
LukasWoodtli: Yes. But again, `pexpect` is not pythonoic as it reqires a string as path argument.

mlasch (approved):

PR #786, opened 2024-06-11
cmake: Allow client related settings from CMake

LukasWoodtli:
The settings for "Client Initiated Bootstrap" and restriction to LwM2M version 1.0 can be now set with CMake options.
These options are only available if the client mode is enabled.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -51,15 +51,17 @@ add_test_variant(
     COMPILE_DEFINITIONS LWM2M_SERVER_MODE
 )
 
-add_test_variant(
-    TARGET_NAME lwm2munittests_client_server_bootstrap_all_formats
-    SOURCE_FILES ${TEST_SOURCES}
-    COMPILE_DEFINITIONS
-        LWM2M_CLIENT_MODE
-        LWM2M_SERVER_MODE
-        LWM2M_BOOTSTRAP_SERVER_MODE
-        LWM2M_SERVER_MODE
-        LWM2M_OLD_CONTENT_FORMAT_SUPPORT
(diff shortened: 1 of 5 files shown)


mlasch:
I tried to build the client with `-DWAKAAMA_CLIENT_LWM2M_V_1_0=ON` and got a build error.

```
/home/marc/repos/wakaama/eclipse-wakaama/data/senml_json.c:30:2: error: #error SenML JSON not supported with LWM2M 1.0
   30 | #error SenML JSON not supported with LWM2M 1.0
      |  ^~~~~
```
This makes sense, maybe `SenML JSON` should be disabled for the client when forcing LwM2M 1.0?

tests/CMakeLists.txt
mlasch, line 1: in line 27 it should look like this.
```suggest
if(WAKAAMA_MODE_CLIENT AND NOT WAKAAMA_MODE_BOOTSTRAP_SERVER)
```

LukasWoodtli:
> I tried to build the client with `-DWAKAAMA_CLIENT_LWM2M_V_1_0=ON` and got a build error.
> 
> ```
> /home/marc/repos/wakaama/eclipse-wakaama/data/senml_json.c:30:2: error: #error SenML JSON not supported with LWM2M 1.0
>    30 | #error SenML JSON not supported with LWM2M 1.0
>       |  ^~~~~
> ```
> 
> This makes sense, maybe `SenML JSON` should be disabled for the client when forcing LwM2M 1.0?

This didn't work before. We just did not have any example or test that combines client version 1.0 with SenML JSON.

We should add more checks to the CMake projects to disallow certain configuration combinations. But for that we need a clean CMake project first.

mlasch (approved):

PR #787, opened 2024-06-12
Update readme

mlasch:
Rework the readme file a bit.

- **Remove layout section**
  This section provides limited use since Github as well as modern IDEs
  have good file browsing built in. The source layout section consumes
  valuable visual space on the Github landing page and distracts from
  other important information.
- **Move CI batch to the top**
  There is no need for a separate section for batches. This nicely fits
  below the project name.

- **Update cmake commands to build examples**
  Update build commands for examples with cmake variables to configure the
  build process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,34 +26,6 @@ This work is dual-licensed under the Eclipse Public License v2.0 and Eclipse Dis
 
 [![Build](https://github.com/eclipse/wakaama/actions/workflows/build_and_test.yaml/badge.svg)](https://github.com/eclipse/wakaama/actions/workflows/build_and_test.yaml)
 
-## Source Layout
-
-    -+- core                   (the LWM2M engine)
-     |
-     +- coap                   (CoAP stack adaptation)
-     |    |
-     |    +- er-coap-13        (Modified Erbium's CoAP engine from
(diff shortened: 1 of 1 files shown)


kwjones:
Thank you for requesting the update to the examples build process.  I'd not worked with Wakaama for many months, and expected the previous cmake/make process to be the same (given the README.md).  Surprise!  Really nice work from all the contributors.  I cannot thank you all enough. Having an "open" client alternative is extremely important. Again, many thanks!

README.md
```diff
@@ -174,10 +144,10 @@ There are some example applications provided to test the server, client and boot
 The following recipes assume you are on a unix like platform and you have cmake and make installed.
 
 ### Server example
- * Create a build directory and change to that.
- * ``cmake [wakaama directory]/examples/server``
- * ``make``
- * ``./lwm2mserver [Options]``
+
+ * ``cmake -S examples/server -B build-server -DWAKAAMA_MODE_SERVER=ON``
```
LukasWoodtli, line 148: Eventually we should use the CMake presets here. But maybe it makes sense to clean up the complete CMake project first.

LukasWoodtli: Same for the other examples.

LukasWoodtli (approved):
Cool! Thanks for the cleanup!

PR #788, opened 2024-06-18
Rename unit test files

mlasch:
Rename the unit test files to match their location in the project. This also triggered a code reformatting from clang-format.


LukasWoodtli (approved):
Thanks for the clean up!

PR #789, opened 2024-06-19
cmake: Allow CoAP related setting from CMake

LukasWoodtli:
The settings regarding raw block-wise transfer and default block size can now be set from CMake.


README.md
```diff
@@ -76,6 +73,13 @@ Wakaama supports additional client related options. These are only available if
  
 Please note: LwM2M version 1.0 is only supported by clients, while servers are backward compatible.
 
+### CoaP Settings
```
mlasch, line 76:
```suggestion
### CoAP Settings
```

README.md
```diff
@@ -76,6 +73,13 @@ Wakaama supports additional client related options. These are only available if
...
+
+- WAKAAMA_RAW_BLOCK1_REQUESTS For low memory client devices where it is not possible to keep a large post or put request in memory to be parsed (typically a firmware write).
```
mlasch, line 78: Should we name this ` WAKAAMA_COAP_RAW_BLOCK1_REQUESTS`? Block transfer is a CoAP feature.

mlasch (changes requested):

README.md
LukasWoodtli: Good idea. This makes the variable name quite long. But it's more consistent with the other variables.

mlasch (approved):

PR #791, opened 2024-06-19
CoAP Packet Deduplication

mlasch:
According to rfc 7252, section "4.5. Message Deduplication", duplicate received packets must be dropped. Also "The recipient SHOULD acknowledge each duplicate copy of a Confirmable message using the same Acknowledgement or Reset message [...]".

Wakaama identifies a duplicate message based on the message id and the session it was received from. Duplicate messages are then processed only once for the first message, subsequent messages are acked with the
corresponding CoAP response code, but not processed further. Duplicate messages are ignored for a predefined time period EXCHANGE_LIFETIME.

### DIFF
--- a/include/liblwm2m.h
+++ b/include/liblwm2m.h
@@ -817,6 +817,8 @@ typedef int (*lwm2m_bootstrap_callback_t)(lwm2m_context_t *contextP, void *sessi
                                           void *userData);
 #endif
 
+typedef struct _coap_msg_dedup_ coap_msg_dedup_t;
+
 struct _lwm2m_context_
 {
 #ifdef LWM2M_CLIENT_MODE
@@ -842,6 +844,7 @@ struct _lwm2m_context_
 #endif
     uint16_t                nextMID;
     lwm2m_transaction_t *   transactionList;
+    coap_msg_dedup_t *message_dedup;
     void *                  userData;
 };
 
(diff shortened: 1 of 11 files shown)


boaks:
If the response has payload, also that payload is intended to be sent back in cases of deduplication, not only the response code.

include/liblwm2m.h
```diff
@@ -817,6 +817,8 @@ typedef int (*lwm2m_bootstrap_callback_t)(lwm2m_context_t *contextP, void *sessi
                                           void *userData);
 #endif
 
+typedef struct _coap_msg_dedup_ coap_msg_dedup_t;
```
LukasWoodtli, line 820: It's not clear if this really belongs here. Probably it does together with the complete struct definition as we need it in the context. But then it should probably be prefixed by `lwm2m_`

mlasch: I had to store the list somewhere, the lwm2m context was the easiest place, for that I needed this forward declaration. The message duplication tracking list is an implementation detail of the CoAP layer and might fit better there. The context structure was available everywhere I needed access though. Problem is, there is no clean separation between LwM2M and CoAP in Wakaama.

LukasWoodtli: I agree that we don't have that nice separation between CoAP and LwM2M. But the forward declaration doesn't work properly together with the typedef.

LukasWoodtli:
Todo: squash fix

LukasWoodtli:
Squash this PR before merging

PR #793, opened 2024-06-20
Add support for send in server side

semhoun:
sbertin-telula patch for send in client was added to wakaama but not the server side.
So I fix the server side part.

### DIFF
--- a/core/internals.h
+++ b/core/internals.h
@@ -369,13 +369,13 @@ typedef struct
 } bs_data_t;
 #endif
 
-typedef enum
-{
+typedef enum {
     LWM2M_REQUEST_TYPE_UNKNOWN,
     LWM2M_REQUEST_TYPE_DM,
     LWM2M_REQUEST_TYPE_REGISTRATION,
     LWM2M_REQUEST_TYPE_BOOTSTRAP,
-    LWM2M_REQUEST_TYPE_DELETE_ALL
+    LWM2M_REQUEST_TYPE_DELETE_ALL,
+    LWM2M_REQUEST_TYPE_SEND,
 } lwm2m_request_type_t;
 
 #ifdef LWM2M_SUPPORT_SENML_CBOR
@@ -598,6 +598,11 @@ lwm2m_server_t * utils_findBootstrapServer(lwm2m_context_t * contextP, void * fr
 #endif
 #if defined(LWM2M_SERVER_MODE) || defined(LWM2M_BOOTSTRAP_SERVER_MODE)
 lwm2m_client_t * utils_findClient(lwm2m_context_t * contextP, void * fromSessionH);
+#endif
+
+// defined in reporting.c
+#if defined(LWM2M_SERVER_MODE) && !defined(LWM2M_VERSION_1_0)
+uint8_t reporting_handleSend(lwm2m_context_t *contextP, void *fromSessionH, coap_packet_t *message);
 #endif
 
 #endif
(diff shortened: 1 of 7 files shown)


core/observe.c
```diff
@@ -1493,4 +1493,33 @@ bool observe_handleNotify(lwm2m_context_t * contextP,
...
+    for (clientP = contextP->clientList; clientP != NULL; clientP = clientP->next) {
+        if (clientP->sessionH == fromSessionH)
+            break;
+    }
+    if (clientP == NULL)
+        return COAP_400_BAD_REQUEST;
+
+    format = utils_convertMediaType(message->content_type);
+
+    if (contextP->monitorCallback != NULL) {
```
mlasch, line 1516: I would rather create a separate callback with an interface like
```
void lwm2m_reporting_set_send_callback(lwm2m_context_t *contextP, lwm2m_result_callback_t callback, void *userData);
```
to report send messages. The `monitorCallback` should only be used for the "Client Registration Interface" and not for the "Information Reporting Interface". Also I would put this in a separate file like `core/reporting.c`.

core/internals.h
```diff
@@ -455,6 +455,9 @@ void observe_clear(lwm2m_context_t * contextP, lwm2m_uri_t * uriP);
 bool observe_handleNotify(lwm2m_context_t * contextP, void * fromSessionH, coap_packet_t * message, coap_packet_t * response);
 void observe_remove(lwm2m_observation_t * observationP);
 lwm2m_observed_t * observe_findByUri(lwm2m_context_t * contextP, lwm2m_uri_t * uriP);
+#ifndef LWM2M_VERSION_1_0
```
mlasch, line 458: It might be useful to add a check for `LWM2M_SERVER_MODE` in addition to `LWM2M_VERSION_1_0`.

core/observe.c
```diff
@@ -1493,4 +1493,33 @@ bool observe_handleNotify(lwm2m_context_t * contextP,
     }
     return true;
 }
+
+#ifndef LWM2M_VERSION_1_0
```
mlasch, line 1497: It might be useful to add a check for `LWM2M_SERVER_MODE` in addition to `LWM2M_VERSION_1_0`.

core/packet.c
```diff
@@ -198,6 +198,11 @@ static uint8_t handle_request(lwm2m_context_t * contextP,
     case LWM2M_REQUEST_TYPE_REGISTRATION:
         result = registration_handleRequest(contextP, &uri, fromSessionH, message, response);
         break;
+#ifndef LWM2M_VERSION_1_0
```
mlasch, line 201: It might be useful to add a check for `LWM2M_SERVER_MODE` in addition to `LWM2M_VERSION_1_0`.

mlasch (changes requested):
Thanks for your contribution. This was still missing indeed, we have an internal patch for which we didn't find the time yet to upstream, but maybe it helps: https://github.com/husqvarnagroup/wakaama/tree/gardena/ml/server-side-send

core/packet.c
semhoun: Is already in a #ifdef LWM2M_SERVER_MODE (line 197)

mlasch: Yes, my bad.

core/observe.c
semhoun: I made the change.

core/observe.c
semhoun: I made the change

semhoun:
I made a separate callback as you suggested

mlasch:
Nice, can you squash the "fix" commits and rebase your branch to the current upstream `main` branch, such that we get rid of the merge commits and have a nice linear commit history?

semhoun:
I rebase to main. I squish all the commits, not only the fix because of the rebase.

mlasch (approved):

mlasch (approved):
Thanks for squashing. Looks good now.

PR #794, opened 2024-06-21
send: Reject content format other than SenML

mlasch:
LwM2M 1.1 send only supports SenML-JSON or SenML-CBOR. So allow only SenML-JSON or SenML-CBOR content format for LWM2M 1.1 send requests.


LukasWoodtli:
Cool! Thanks!

LukasWoodtli (approved):

PR #795, opened 2024-06-21
send: Return not found if client is NULL

mlasch:
Return 4.04 not found in case the provided client is not valid.

### DIFF
--- a/core/reporting.c
+++ b/core/reporting.c
@@ -36,7 +36,7 @@ uint8_t reporting_handleSend(lwm2m_context_t *contextP, void *fromSessionH, coap
             break;
     }
     if (clientP == NULL)
-        return COAP_400_BAD_REQUEST;
+        return COAP_404_NOT_FOUND;
 
     format = utils_convertMediaType(message->content_type);
 


core/reporting.c
```diff
@@ -36,7 +36,7 @@ uint8_t reporting_handleSend(lwm2m_context_t *contextP, void *fromSessionH, coap
             break;
     }
     if (clientP == NULL)
-        return COAP_400_BAD_REQUEST;
+        return COAP_404_NOT_FOUND;
```
rettichschnidi, line 39: What is the reasoning for this change?

Reading up on [404](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.4) ("not find a current representation *for the target resource*"),  I get the feeling that 400 is more appropriate.

sbernard31: This is about Send Operation ? 

For error code in LWM2M, you can also look at : [LWM2M-v1.1.1@transportTable: 6.7.-4 Response Codes: Information Reporting Interface](https://www.openmobilealliance.org/release/LightweightM2M/V1_1_1-20190617-A/HTML-Version/OMA-TS-LightweightM2M_Transport-V1_1_1-20190617-A.html#Table-67-4-Response-Codes-Information-Reporting-Interface)

And they say : 
>**Send :** 
>2.04 Changed : 	"Send" operation completed successfully
>4.00 Bad Request : 	Undetermined error occurred
>4.04 Not Found : 	Reported Object was not registered to the LwM2M Server

In Leshan, I think we send : 
`BAD_REQUEST` this error message  "no registration found" when we receive Send request from a device which is currently not registered)

mlasch:
Closed due to review feedback. Thanks for the input.

PR #796, opened 2024-06-28
Move the connections implementations to own folders

LukasWoodtli:
This leads to a cleaner project structure and forms the basis for a more modular project.


mlasch:
Good idea to move the connection implementations out of the examples directory. I would not put the connection implementations in an "addon" directory. Typically an application (or example) will use one of the connections making it not really an addon. I suggest to put the networking layer in a directory called something like "transport" which would then contain "udp" and "tinydtls".

This to me personally looks much cleaner.
```
coap
core
...
transport
├── tinydtls
└── udp
```

mlasch (approved):

PR #803, opened 2024-08-15
CMake options for connection layer

LukasWoodtli:
The connection layer can now be selected with a CMake cache variable.

Some modularization of the CMake project was done in the process.


mlasch:
Looks sensible to me.

PR #806, opened 2024-08-29
CMake: platform layer and CLI library

LukasWoodtli:


README.md
```diff
@@ -100,6 +100,20 @@ The logging infrastructure can be configured with CMake cache variables (e.g. `c
 
 If `NONE` is chosen, the user of Wakaama needs to implement a custom transport layer. Check the available implementations for more information.
 
+### Platform
+
+- WAKAAMA_PLATFORM: Select the implementation of the platform abstraction layer One of:
```
rettichschnidi, line 105:
```suggestion
- WAKAAMA_PLATFORM: Select the implementation of the platform abstraction layer, one of:
```

wakaama.cmake
```diff
@@ -27,52 +27,72 @@ option(WAKAAMA_DATA_OLD_CONTENT_FORMAT "Support the deprecated content format va
...
-             LOG_DISABLED
-             16
-             32
-             64
-             128
-             256
-             512
-             1024
-)
+if(NOT WAKAAMA_COAP_DEFAULT_BLOCK_SIZE)
```
rettichschnidi, line 30: Not sure what exactly the issue was before, but set() has been greatly improved [starting with CMake 3.21](https://cmake.org/cmake/help/latest/policy/CMP0126.html#policy:CMP0126). I suggest we try updating cmake_minimum_required() to (at least) CMake 3.21 and check this again.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.20)
+cmake_minimum_required(VERSION 3.21)
```
rettichschnidi, line 1: While this, and all other such changes, is fine, file `examples/CMakeLists.txt` should be updated too I think.

rettichschnidi (approved):

CMakeLists.txt
LukasWoodtli: Somehow I missed that one. Thanks for the hint!

PR #809, opened 2024-09-09
Remove code duplication in client example

LukasWoodtli:
A lot of code was copy-and-pasted several times and functions were too long.

This is done to make Sonar Cloud happy. It considers moved files to be new code and complains about not beeing conformant.

But it's anyway good to have some clean-ups.

### DIFF
--- a/examples/client/object_connectivity_moni.c
+++ b/examples/client/object_connectivity_moni.c
@@ -86,6 +86,19 @@ typedef struct
     int linkUtilization;
 } conn_m_data_t;
 
+static void reduce_single_instance(lwm2m_data_t *const dataP, lwm2m_data_t **const subTlvP, size_t *const count) {
+    if (dataP->type == LWM2M_TYPE_MULTIPLE_RESOURCE) {
+        (*count) = dataP->value.asChildren.count;
+        (*subTlvP) = dataP->value.asChildren.array;
+    } else {
+        (*count) = 1; // reduced to 1 instance to fit in one block size
+        (*subTlvP) = lwm2m_data_new((*count));
+        for (size_t i = 0; i < (*count); i++)
+            (*subTlvP)[i].id = i;
+        lwm2m_data_encode_instances((*subTlvP), (*count), dataP);
+    }
+}
+
 static uint8_t prv_set_value(lwm2m_data_t * dataP,
                              conn_m_data_t * connDataP)
 {
@@ -102,18 +115,7 @@ static uint8_t prv_set_value(lwm2m_data_t * dataP,
 
     case RES_M_AVL_NETWORK_BEARER:
     {
-        if (dataP->type == LWM2M_TYPE_MULTIPLE_RESOURCE)
-        {
-            count = dataP->value.asChildren.count;
-            subTlvP = dataP->value.asChildren.array;
-        }
-        else
-        {
-            count = 1; // reduced to 1 instance to fit in one block size
-            subTlvP = lwm2m_data_new(count);
-            for (i = 0; i < count; i++) subTlvP[i].id = i;
-            lwm2m_data_encode_instances(subTlvP, count, dataP);
-        }
+        reduce_single_instance(dataP, &subTlvP, &count);
 
         for (i = 0; i < count; i++)
         {
@@ -141,18 +143,7 @@ static uint8_t prv_set_value(lwm2m_data_t * dataP,
 
     case RES_M_IP_ADDRESSES:
     {
-        if (dataP->type == LWM2M_TYPE_MULTIPLE_RESOURCE)
-        {
-            count = dataP->value.asChildren.count;
-            subTlvP = dataP->value.asChildren.array;
-        }
-        else
-        {
-            count = 1; // reduced to 1 instance to fit in one block size
-            subTlvP = lwm2m_data_new(count);
-            for (i = 0; i < count; i++) subTlvP[i].id = i;
-            lwm2m_data_encode_instances(subTlvP, count, dataP);
-        }
+        reduce_single_instance(dataP, &subTlvP, &count);
 
         for (i = 0; i < count; i++)
         {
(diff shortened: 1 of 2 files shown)


examples/client/object_server.c
```diff
@@ -224,6 +224,50 @@ static uint8_t prv_get_value(lwm2m_data_t * dataP,
...
+            if (resList[i] == id) {
+                *nbRes -= 1;
+                memmove(&resList[i], &resList[i + 1], (*nbRes - i) * sizeof(resList[i]));
+                break;
+            }
+        }
+    }
+}
+
+static void setup_resources(lwm2m_data_t *const *dataArrayP, const uint16_t *resList, int nbRes) {
```
rettichschnidi, line 240:
```suggestion
static void setup_resources(lwm2m_data_t *const *dataArrayP, const uint16_t *const resList, const int nbRes) {
```

rettichschnidi: Also, while on it, would change `int nbRes` to `size_t nbRes` and clean up the caller sites. While extra work, this seems quite little effort.

Feel free to ignore however, as this would also make a perfect follow up PR.

examples/client/object_server.c
```diff
@@ -224,6 +224,50 @@ static uint8_t prv_get_value(lwm2m_data_t * dataP,
     }
 }
 
+static void remove_optional_resource(uint16_t *resList, int val, int id, int *nbRes) {
```
rettichschnidi, line 227: I know this is a matter of taste, but I think using early exits and therefore keeping the happy path as little nested as possible, helps readability:

```
static void remove_optional_resource(uint16_t *const resList, int val, int id, int *nbRes) {
    if (val >= 0) {
        /* some explanation on why quitting here */
        return;
    }

    for (int i = 0; i < *nbRes; i++) {
        if (resList[i] == id) {
            *nbRes -= 1;
            memmove(&resList[i], &resList[i + 1], (*nbRes - i) * sizeof(resList[i]));
            break;
        }
    }
}
```

rettichschnidi, line 227: If we want to follow trough with constification:
```suggestion
static void remove_optional_resource(uint16_t *const resList, const int val, const int id, int *const nbRes) {
```

rettichschnidi, line 227: It seems to me like `id` will always be positive and small:
```suggestion
static void remove_optional_resource(uint16_t *resList, int val, const uint_fast8_t id, int *nbRes) {
```

examples/client/object_server.c
```diff
@@ -224,6 +224,50 @@ static uint8_t prv_get_value(lwm2m_data_t * dataP,
...
+    }
+}
+
+static void setup_resources(lwm2m_data_t *const *dataArrayP, const uint16_t *resList, int nbRes) {
+    for (int i = 0; i < nbRes; i++) {
+        (*dataArrayP)[i].id = resList[i];
+    }
+}
+
+static void remove_all_optional_resources(uint16_t *resList, server_instance_t *targetP, int *nbRes) {
```
rettichschnidi, line 246: Constification + using a more fitting data type:
```suggestion
static void remove_all_optional_resources(uint16_t *const resList, const server_instance_t *const targetP,
                                          size_t *const nbRes) {
```

examples/client/object_server.c
```diff
@@ -423,111 +356,13 @@ static uint8_t prv_server_discover(lwm2m_context_t *contextP,
 
 #ifndef LWM2M_VERSION_1_0
         /* Remove optional resources that don't exist */
```
rettichschnidi, line 358: Would use this this as a starting point to document the newly created function, not "waste" it here.

examples/client/object_connectivity_moni.c
```diff
@@ -86,6 +86,19 @@ typedef struct
     int linkUtilization;
 } conn_m_data_t;
 
+static void reduce_single_instance(lwm2m_data_t *dataP, lwm2m_data_t **subTlvP, size_t *count) {
```
rettichschnidi, line 89: SCNR:
```suggestion
static void reduce_single_instance(lwm2m_data_t *const dataP, lwm2m_data_t **const subTlvP, size_t *const count) {
```

examples/client/object_server.c
```diff
@@ -224,6 +224,50 @@ static uint8_t prv_get_value(lwm2m_data_t * dataP,
...
+    remove_optional_resource(resList, targetP->registrationFailureBlock, LWM2M_SERVER_REG_FAIL_BLOCK_ID, nbRes);
+    remove_optional_resource(resList, targetP->bootstrapOnRegistrationFailure, LWM2M_SERVER_REG_FAIL_BOOTSTRAP_ID,
+                             nbRes);
+    remove_optional_resource(resList, targetP->communicationRetryCount, LWM2M_SERVER_COMM_RETRY_COUNT_ID, nbRes);
+    remove_optional_resource(resList, targetP->communicationRetryTimer, LWM2M_SERVER_COMM_RETRY_TIMER_ID, nbRes);
+    remove_optional_resource(resList, targetP->communicationSequenceDelayTimer, LWM2M_SERVER_SEQ_DELAY_TIMER_ID, nbRes);
+    remove_optional_resource(resList, targetP->communicationSequenceRetryCount, LWM2M_SERVER_SEQ_RETRY_COUNT_ID, nbRes);
+}
+
+static uint8_t prv_get_all_values(const int *numDataP, lwm2m_data_t *const *dataArrayP, server_instance_t *targetP) {
```
rettichschnidi, line 258:
```suggestion
static uint8_t prv_get_all_values(const int *numDataP, lwm2m_data_t *const *dataArrayP, server_instance_t *const targetP) {
```

rettichschnidi (changes requested):
Nice commits, good content, yet I have some nitpicks. :D

Also, I do understand this is just a refactoring, but if it would be easy for you do to do, adding some explanation (especially about the return values) of those new functions would have been helpful for me.

examples/client/object_server.c
LukasWoodtli: Resource ID is 16 bit.

rettichschnidi (approved):

PR #812, opened 2024-09-29
Lw/static lib

LukasWoodtli:
Create a static library for Wakaama.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -57,4 +57,5 @@ jobs:
         tools/ci/run_ci.sh \
           --run-clean \
           --run-build \
-          --source-directory examples/${{ matrix.example }}
+          --source-directory examples/${{ matrix.example }} \
+          --build-directory wakaama-build-${{ matrix.example }}
--- a/tools/ci/run_ci.sh
+++ b/tools/ci/run_ci.sh
@@ -29,6 +29,7 @@ OPT_SANITIZER=""
 OPT_SCAN_BUILD=""
 OPT_SONARQUBE=""
 OPT_SOURCE_DIRECTORY="${REPO_ROOT_DIR}"
+OPT_BUILD_DIRECTORY="build-wakaama"
 OPT_TEST_COVERAGE_REPORT=""
 OPT_VERBOSE=0
 OPT_WRAPPER_CMD=""
@@ -58,6 +59,9 @@ Options:
                             (BINARY: defaults to ${OPT_CLANG_FORMAT})
   --source-directory PATH   Configure CMake using PATH instead of the
                             repositories root directory.
+  --build-directory PATH    Configure CMake using PATH as the build directory.
+                            Defaults to 'build-wakaama'. Builds from CMake
+                            presets are placed into 'build-presets' by default.
   --sanitizer TYPE          Enable sanitizer
                             (TYPE: address leak thread undefined)
   --scan-build BINARY       Enable Clang code analyzer using specified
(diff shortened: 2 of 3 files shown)


wakaama.cmake
```diff
@@ -323,6 +323,25 @@ function(target_sources_shared target)
     target_include_directories(${target} PUBLIC ${WAKAAMA_EXAMPLE_SHARED_DIRECTORY})
 endfunction()
 
+# Static library that users of Wakaama can link against
```
rettichschnidi, line 326: I do not understand why having a static library is a good thing to have in the first place. AFAIR, the Wakaama way of doing things is "here are the sources, and there are 100x configuration parameters, pick yours and build it".

.github/workflows/build.yaml
```diff
@@ -57,4 +57,5 @@ jobs:
         tools/ci/run_ci.sh \
           --run-clean \
           --run-build \
-          --source-directory examples/${{ matrix.example }}
+          --source-directory examples/${{ matrix.example }} \
+          --build-directory wakaama-build-${{ matrix.example }}
```
rettichschnidi, line 61: I can not comment on the commit message itself, so I do it here instead: What is the goal of this change? Will it allow e.g. for multiple builds ending up in the same CI artifact? Or is this just an improvement for local development?

wakaama.cmake
LukasWoodtli: Wakaama became difficult to build. Where to set which defines and compiler flags, what files to include into the build, which include directories to use... That's after all, that's why we have build systems and software libraries. You will still need to pick your configuration parameters, but you don't need to care about the rest.

.github/workflows/build.yaml
LukasWoodtli: This way you can build the examples separately without overwriting previous builds. It's useful for local development. But it will mainly be used when we refactor the examples to be individual projects, that depend on Wakaama as a 3rd-party.
Currently, the examples are tightly coupled to the rest of the project.

wakaama.cmake
LukasWoodtli: It's almost impossible to modularize the project without the use of a capable build system. Currently, the project is very monolithic and hard to maintain.

.github/workflows/build.yaml
rettichschnidi: I think this comment would be good to have in the commit message.

wakaama.cmake
rettichschnidi: I think this comment would be good to have in the commit message and/or in wakaama.cmake.

rettichschnidi (approved):
I think it would be beneficial to have your clarifying comments to my concerns to be stored in the Git commit messages and/or as comments in the sources.

But I am not insisting on this and therefore approving the PR.

tools/ci/run_ci.sh
```diff
@@ -58,6 +59,8 @@ Options:
                             (BINARY: defaults to ${OPT_CLANG_FORMAT})
   --source-directory PATH   Configure CMake using PATH instead of the
                             repositories root directory.
+  --build-directory PATH    Configure CMake using PATH as the build directory.
+                            Defaults to 'build-wakaama'.
```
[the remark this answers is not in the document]
LukasWoodtli: True. Maybe add a comment about the build-presets.

.github/workflows/build.yaml
LukasWoodtli: You are right! I'll add some clarification to the commit.

mlasch:
It might be worthwhile to give some instructions in the readme on how to build and use the static library.

mlasch (approved):

PR #819, opened 2024-11-10
readme: Add OpenSSF Best Practices badge

rettichschnidi:
While we are not perfect, this gives us a clear list of tasks to do in order to become better.

While I am listed as owner of the badge, everyone with write access to the repository should be able to update its data.

FYI: I'll transfer the ownership of the badge to @LukasWoodtli. But this is [a manual process](https://github.com/coreinfrastructure/best-practices-badge/issues/2126) and I have to (?) do this using my private e-mail address, thus not during work time.

image

Profile: https://www.bestpractices.dev/en/projects/9686/


LukasWoodtli (commented):
Cool! Thanks!

LukasWoodtli (approved):

PR #824, opened 2024-11-14
ci: Use CodeChecker as static analysis tool

LukasWoodtli:
CodeChecker combines multiple static code analysis tools including `clang-sa`. It also provides the possibility to check against a baseline to report only newly introduced issues.
Moreover it is simpler to use.
It could eventually replace `clang-sa` completely in the pipeline.


LukasWoodtli:
Works locally but doesn't work on GH. Don't know why.

PR #827, opened 2024-11-15
ci: Enforce pinned pip dependencies

rettichschnidi:
This should give us a 10/10 OpenSSF rating for pinned dependencies.

It has been done all by hand, trough trial and error. Feels like there should be some better tooling for this out there...


LukasWoodtli (approved):
Cool!

rettichschnidi:
> Cool!

Let's see if we think the same a few months down the road. :)

PR #834, opened 2024-11-26
Fix repository name

rettichschnidi:
While https://github.com/eclipse/wakaama redirects to https://github.com/eclipse-wakaama/wakaama, the later is what should be used nowadays.


LukasWoodtli:
nit: We didn't use the prefix 'global' yet. Maybe just use 'doc'?

LukasWoodtli (approved):

PR #841, opened 2025-01-04
Lw/safer string functions

LukasWoodtli:
Provide safer variants of `strnlen` and `strncpy` which check the provided bounds of the char arrays.

### DIFF
--- a/core/utils.h
+++ b/core/utils.h
@@ -47,5 +47,15 @@ lwm2m_server_t *utils_findBootstrapServer(lwm2m_context_t *contextP, void *fromS
 #if defined(LWM2M_SERVER_MODE) || defined(LWM2M_BOOTSTRAP_SERVER_MODE)
 lwm2m_client_t *utils_findClient(lwm2m_context_t *contextP, void *fromSessionH);
 #endif
+/** A safer version of `strlen`. Finds the number of chars (bytes) of the given string up to a given max. length.
+ *
+ * The behaviour is undefined if the maximum length value is bigger than the actually reserved memory behind the @a str
+ * buffer.
+ *
+ * @param str The null-terminated string to find the length of.
+ * @param max_size Max size of string.
+ * @return The size of the string if it is smaller or equal to max_size. Otherwise max_size.
+ */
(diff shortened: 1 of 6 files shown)


core/utils.h
```diff
@@ -26,6 +26,16 @@ lwm2m_media_type_t utils_convertMediaType(coap_content_type_t type);
 uint8_t utils_getResponseFormat(uint8_t accept_num, const uint16_t *accept, int numData, const lwm2m_data_t *dataP,
                                 bool singleResource, lwm2m_media_type_t *format);
 int utils_isAltPathValid(const char *altPath);
+/** Copy a string.
+ *
+ * Use utils_strncpy if possible!
+ * This is less safe than utils_strncpy.
+ *
+ * @param buffer The destination buffer
+ * @param length The mex. size of the destination buffer
```
rettichschnidi, line 35: a) Typo
b) "Max size of teh buffer" feels like a wrong description.

Maybe this?
```suggestion
 * @param length The max number of bytes to be written to in the destination buffer
```

core/utils.h
```diff
@@ -47,5 +57,25 @@ lwm2m_server_t *utils_findBootstrapServer(lwm2m_context_t *contextP, void *fromS
...
+ *
+ * @param str The null-terminated string to find the length of.
+ * @param max_size Max size of string.
+ * @return The size of the string if it is smaller or equal to max_size. Otherwise max_size.
+ */
+size_t utils_strnlen(const char *str, size_t max_size);
+
+/** A safer version of `strncpy`. Copies at most src_size bytes to dest, but checks for available space.
+ *
+ * If dest is not NULL and dest_size is not 0 the destination buffer is always terminated with a '\0'.
```
rettichschnidi, line 72:
```suggestion
 * If dest is not NULL and dest_size is not 0, then the destination buffer is always terminated with '\0'.
```

rettichschnidi (commented):
Some quick, initial feedback. Will do a more thorough review once I can spare some time.

rettichschnidi (approved):

PR #842, opened 2025-01-22
coap: Cleanup transactions when client is unregistered

LukasWoodtli:
All pending transactions of a just unregistered client must be removed


mlasch:
Thanks for the upstreaming effort.

mlasch (approved):

PR #846, opened 2025-01-28
send: return not found on unknown session

LukasWoodtli:
Previously 'bad request' was returned.


mlasch:
This is a duplicate of #795. Reading up the comments of #795 again, I think we should drop this change and rather add a comment to the section to clarify the rationale and avoid confusion in the future.

PR #848, opened 2025-01-30
Lw/fix GitHub run on arch le

LukasWoodtli:


LukasWoodtli:
Doesn't work: gcc has bug (SEGFAULT)

PR #851, opened 2025-02-05
cmake: Add option for disabling unit tests

LukasWoodtli:
Building the unit tests can now be disabled with a CMake cache variable. This is useful on platforms where `CUnit` is not available or when working on an embedded platform.


nmatas-tdk (approved):
Looks good to me

PR #858, opened 2025-03-17
cmake: Introduce configuration option for max. package size

LukasWoodtli:
The package size needs to be bigger than the block size as it also needs to hold the CoAP header and options.


mlasch (approved):
Looks about right, tested the server with a Zephyr LwM2M client.

PR #861, opened 2025-04-03
CoAP: Fix empty ack

sarchey:
CoAP: Fix the implementation of the empty ACK and separate response mechanism

-  Fix received message not correct processed
-  Fix timeout for separate response

### DIFF
--- a/coap/er-coap-13/er-coap-13.h
+++ b/coap/er-coap-13/er-coap-13.h
@@ -48,7 +48,7 @@
 
 #define COAP_DEFAULT_MAX_AGE                 60
 #define COAP_RESPONSE_TIMEOUT                2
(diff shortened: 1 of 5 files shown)


coap/er-coap-13/er-coap-13.h
```diff
@@ -48,6 +48,7 @@
 
 #define COAP_DEFAULT_MAX_AGE                 60
 #define COAP_RESPONSE_TIMEOUT                2
+#define COAP_SEPARATE_TIMEOUT                15
```
mlasch, line 51: It would be great to make this configurable from cmake.

sarchey: done

mlasch: Nice, thanks!

PR #862, opened 2025-04-11
Fix: Check memory allocation before memset operation

ashishmac7:
- Moved the NULL check after memory allocation to occur before the memset call, preventing potential undefined behavior if the allocation fails.
- Ensures safe memory handling and avoids crashes due to dereferencing a NULL pointer.

### DIFF
--- a/core/registration.c
+++ b/core/registration.c
@@ -1658,8 +1658,8 @@ static lwm2m_client_object_t *prv_decodeRegisterPayload(uint8_t *payload, size_t
             if (objectP == NULL)
             {
                 objectP = (lwm2m_client_object_t *)lwm2m_malloc(sizeof(lwm2m_client_object_t));
-                memset(objectP, 0, sizeof(lwm2m_client_object_t));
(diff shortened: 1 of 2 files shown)


mlasch:
Hi @ashishmac7 thanks for your contribution. There is a line limit for the git commit message which is enforced by the CI. Can you adjust your commit message accordingly?

```
Run tools/ci/run_ci.sh --branch-target origin/main --run-gitlint

Commit c236965e37:
3: B1 Line exceeds max length (144>80): "- Moved the NULL check after memory allocation to occur before the memset call, preventing potential undefined behavior if the allocation fails."
4: B1 Line exceeds max length (86>80): "- Ensures safe memory handling and avoids crashes due to dereferencing a NULL pointer."
```

ashishmac7:
Hi @mlasch , thanks for pointing that out.

I've updated the commit message to conform to the line length limits enforced by the CI.
Let me know if there's anything else you'd like me to address.

Thanks!

PR #863, opened 2025-04-23
registration: Allow media type to be empty

mlasch:
This reverts "fix: Media type for registration must be CoRE Link", commit 7aa5e9e0dc99db0db5b71f0a30f3f2cbd322c229.

For registration updates (POST) and de-registration operations (DELETE) the content type is not mandatory. Sending these requests without CoRE Link media type should not result in a 4.00 Bad Request error.


LukasWoodtli:
LGTM. But fix tests, please.

mlasch:
Thanks for the feedback, fixed the tests

PR #864, opened 2025-04-26
fix a memory leak bug about a registion update request without a reply

xiaoshan1234:
fix a memory leak bug about a registion update request without a reply
refer to https://github.com/eclipse-wakaama/wakaama/issues/811

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 [![OpenSSF Scorecard](https://api.scorecard.dev/projects/github.com/eclipse-wakaama/wakaama/badge)](https://scorecard.dev/viewer/?uri=github.com/eclipse-wakaama/wakaama)
 
+[![OpenSSF Best Practices](https://www.bestpractices.dev/projects/9686/badge)](https://www.bestpractices.dev/projects/9686)
+
 Wakaama (formerly liblwm2m) is an implementation of the Open Mobile Alliance's LightWeight M2M
 protocol (LWM2M).
 


core/registration.c
```diff
@@ -857,10 +857,13 @@ static void prv_handleRegistrationUpdateReply(lwm2m_context_t * contextP,
             LOG_ARG_DBG("%d Registration update failed", dataP->server->shortID);
         }
     }
-    if (packet != NULL && packet->code != COAP_231_CONTINUE)
-    {
-        lwm2m_free(dataP->payload);
-        transaction_free_userData(contextP, transacP);
+    if (packet != NULL && packet->code == COAP_231_CONTINUE) {
```
mlasch, line 860: If I understand your code correctly, you want to free the user data payload only if `transaction_free_userData` was able to find the actual transaction in the global list and was able to free the user data. Correct me if I'm wrong, but if the transaction cannot be found anymore it should have been freed correctly before. So this might prevent a double free in this case?

Why not inverting the logic to get rid of the empty block after the if-statement?
```suggestion
    if (packet != NULL && packet->code != COAP_231_CONTINUE) {
        uint8_t *payload = dataP->payload;
        if (transaction_free_userData(contextP, transacP)) {
            lwm2m_free(payload);
        }
    }
```

Also can you rebase your change instead of merging? This would remove the merge commit from your branch.

xiaoshan1234: Your understanding is correct, but that's not the point.

I reversed the logic to still be able to free userdata when packet is NULL.

xiaoshan1234: The key is that when packet==NULL, my code and the original code execute different logic.

xiaoshan1234: https://github.com/eclipse-wakaama/wakaama/pull/867/commits
this is a new pull request with a rebase

PR #882, opened 2026-04-13
fix: guard against empty string in prv_parseItem to prevent OOB read

Eun0us:
Hey,

In `prv_parseItem()` (data/senml_cbor.c), when a CBOR string label
is decoded with length 0, the code accesses:

    data.value.asBuffer.buffer[data.value.asBuffer.length - 1]

which underflows to `buffer[-1]` — a read at an arbitrary address.
This causes a SIGSEGV on crafted SenML-CBOR input via `lwm2m_data_parse()`.

The fix is a simple length guard before the buffer access.

Cheers


data/senml_cbor.c
```diff
@@ -113,6 +113,9 @@ static int prv_parseItem(const uint8_t *buffer, size_t bufferLen, senml_record_t
                 label = SENML_CBOR_OBJECT_LINK_LABEL;
             } else {
                 /* Label ending in _ must be supported or generate error. */
```
LukasWoodtli, line 115: I think this comment belong to the second `if` statement, which was already there before.

LukasWoodtli (changes requested):
Thanks. LGTM, aside from the code comment.

PR #886, opened 2026-05-13
ci: Remove sonarcloud scan

rettichschnidi:
The used action is no longer available [1], and moving to the successor would require efforts nobody is willing to invest.

Therefore, removing the action seems like the best course of action.

https://github.com/SonarSource/sonarqube-github-c-cpp/blob/master/README.md


LukasWoodtli (approved):
Thanks!

PR #888, opened 2026-05-20
utils: Add `strlen` function that can be limited in max. length

LukasWoodtli:
The function searches for the terminating `NULL` character in the given string.
It can be configured to search only up to a given maximun length.


rettichschnidi:
What is the goal of this?

LukasWoodtli:
Giving the user the possibility to define a max. string length and therefore possibly improve security and reliability.

